PR #8, opened 2022-05-11
Feature/cc 0004 navigation bar

Jzeva:
We created a navbar and tested it to make sure it works. The NavigationBar can be found in components/NavBar directory.

UI 
image


src/styles/components/navbarBtnStyle.tsx
```diff
@@ -0,0 +1,19 @@
+export default {
+  // Styles for the base style
+  baseStyle: {},
```
kritomomo, line 3: If there is no value, I think it can be deleted

src/styles/components/navbarBtnStyle.tsx
```diff
@@ -0,0 +1,19 @@
...
+    navbarIconBtn: {
+      bg: "transparent",
+      color: "white",
+      fontSize: "xl",
+      _hover: { bg: "#496A80" },
+    },
+  },
+  // The default `size` or `variant` values
+  defaultProps: {
+    size: "sm",
```
kritomomo, line 17: if you just wanna the navbar button to be "sm", just move this to nvabarIconBtn

src/styles/theme.ts
```diff
@@ -18,12 +19,23 @@ const styles = {
...
+    secondary: "#40B484",
   },
 };
 
-const theme = extendTheme(styles, config, colors);
+const components = {
+  Button: { ...Navbar },
+};
+
+const overrides = {
```
kritomomo, line 31: I am ok with this one, but what the difference between
`const theme = extendTheme({ colors: {} })` and 
`const overrides = {colors: {}}. const theme = extendTheme(overrides)`
😂

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+          <IconButton
+            aria-label="Revert edit"
+            icon={<RiArrowGoBackLine />}
+            variant="navbarIconBtn"
+          />
+          <IconButton
+            aria-label="Forward edit"
+            icon={<RiArrowGoForwardLine />}
+            variant="navbarIconBtn"
+            color="#72818B"
```
kritomomo, line 31: The logic here is when you have actions that require undo or redo, the icon will become #FFFFFF, but at the beginning, the color of both buttons is "#72818B", Of course it's okay to write this now.

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,70 @@
+import { Flex, Button, Text, IconButton, Grid, Menu, MenuButton } from "@chakra-ui/react";
+import { FaRegUser } from "react-icons/fa";
+import { FiDownload } from "react-icons/fi";
+import { HiOutlineShoppingBag } from "react-icons/hi";
+import { IoIosArrowBack } from "react-icons/io";
+import { RiArrowGoBackLine, RiArrowGoForwardLine } from "react-icons/ri";
+import { BiStar, BiPencil } from "react-icons/bi";
+
+const NavigationBar = (): JSX.Element => {
```
kritomomo, line 9: https://stackoverflow.com/questions/58656026/what-is-the-difference-between-react-fc-and-jsx-element

Not sure about this one, need to check with @devharris7

src/components/NavBar/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+        <IconButton aria-label="Download design" icon={<FiDownload />} variant="navbarIconBtn" />
+        <IconButton aria-label="Order" icon={<HiOutlineShoppingBag />} variant="navbarIconBtn" />
+        <Button
+          bg="brand.secondary"
+          color="white"
+          width="130px"
+          height="45px"
+          marginLeft="10px"
+          fontSize="xl"
+          data-testid="share"
```
kritomomo, line 61: We can use getByText

kritomomo (commented):
⛽️⛽️⛽️⛽️

tatoMa:
As our newer version of UI has been created. The Navigation height is assigned as 64px. Please have a check the root node (which is a grid) and define the element height to 64px. 
After defining the height to a fixed number, you might also need to modify the padding all to only padding x.
image

tatoMa:
Final modification for me. The Hover state of this share button is too white therefore it reduced the color contrast ratio to only 1.23 (which means hard to read). Please find a color or alpha level to match the same button on the Canva website. 
image
image

kritomomo (approved):
LGTM

tatoMa (approved):
LGTM

PR #12, opened 2022-05-15
feat: cc-0005 navigate to the previous page

Jzeva:
We add the navigation function for the “Home” button which can lead users to the [](https://courtcanva.com/).

Screenshot:
image
image


src/components/NavBar/index.tsx
```diff
@@ -5,14 +5,17 @@ import { HiOutlineShoppingBag } from "react-icons/hi";
...
+import Link from "next/link";
 
 const NavigationBar = () => {
   return (
     <Grid templateColumns="repeat(3, 1fr)" bg="brand.primary" p={4}>
       <Flex alignItems="center">
-        <Button leftIcon={<IoIosArrowBack />} pl="0" variant="navbarIconBtn">
-          Home
-        </Button>
+        <Link href="https://courtcanva.com/" passHref>
```
LawrenceLiao, line 14: I suggest extracting this URL as a constant so that we can dynamically change it according to different environments.  If I remember correctly, this URL is of the landing page in the Production environment. We should keep all environments independent, so UAT or even other environments should have their own landing page instead of sharing the same one with PROD.

src/constants/homePage.ts
```diff
@@ -0,0 +1,3 @@
+const HOME_PAGE_LINK = "https://courtcanva.com/";
+
+export default HOME_PAGE_LINK;
```
devharris7, line 3: eof

src/components/NavBar/index.tsx
```diff
@@ -5,14 +5,18 @@ import { HiOutlineShoppingBag } from "react-icons/hi";
...
 
 const NavigationBar = () => {
   return (
     <Grid templateColumns="repeat(3, 1fr)" bg="brand.primary" p={4}>
       <Flex alignItems="center">
-        <Button leftIcon={<IoIosArrowBack />} pl="0" variant="navbarIconBtn">
-          Home
-        </Button>
+        <Link href={HOME_PAGE_LINK} passHref>
+          <Button leftIcon={<IoIosArrowBack />} pl="0" variant="navbarIconBtn">
```
devharris7, line 16: what's `pl`?

tatoMa, line 16: Maybe consider using padding x as 8px and try if it looks better. Cos this padding left makes the home button weird on itss event shape. Or do you have a firm reason for using pl 0?
image

Jzeva: "pl" means padding-left.

src/constants/homePage.ts
```diff
@@ -0,0 +1,3 @@
+const HOME_PAGE_LINK = "https://courtcanva.com/";
```
kritomomo, line 1: It seems that this url only belongs to the navbar. Why create a folder outside the navbar file

Jzeva: Because we were considering  traditionally setting these kind of URLs as constants would be more maintainable and were reusable for being used in other components possibly in the future. Lawrence seems to agree with this way in the first comment.

kritomomo: Ok～，just change the name to index.ts

src/components/NavBar/index.tsx
```diff
@@ -6,7 +6,7 @@ import { IoIosArrowBack } from "react-icons/io";
 import { RiArrowGoBackLine, RiArrowGoForwardLine } from "react-icons/ri";
 import { BiStar, BiPencil } from "react-icons/bi";
 import Link from "next/link";
-import HOME_PAGE_LINK from "@/constants/homePage";
+import HOME_PAGE_LINK from "@/constants/index";
```
Fuchih, line 9: there is no nedd to add /index in your path

Fuchih (approved):
looks good to me

kritomomo (approved):
LGTM

src/constants/index.ts
```diff
@@ -0,0 +1,3 @@
+const HOME_PAGE_LINK = "https://courtcanva.com/";
```
devharris7, line 1: constants/colors.ts
constants/zIndex.ts

src/constants.ts

export const HOME_PAGE_LINK = "https://courtcanva.com/";
export const 404_PAGE_LINK = "https://courtcanva.com/404";

import {HOME_PAGE_LINK}

PR #15, opened 2022-05-16
fix: testing error related to svg caused by nextjs overwrite svg stud…

tatoMa:
This is the workaround way of fix for testing errors when mocking SVG files.

Basically, Nextjs overwrite the jest config with its own code so that the jest config is modified in the wrong order. This is a workaround to modified back the config with the prioritized SVG stud module and move it to the first of the mapper order with the created custom mock file.

As a result, all SVG testing will be normal with no errors.

### DIFF
--- a/__mocks__/svgMock.js
+++ b/__mocks__/svgMock.js
@@ -1,0 +1,11 @@
+import React from "react";
+
+const svgMock = (...props) => {
+  const svg = React.createElement("svg", {
+    fill: "currentColor",
+    xmlns: "http://www.w3.org/2000/svg",
+    ...props[0],
+  });
+  return svg;
+};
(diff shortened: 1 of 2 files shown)


kritomomo (approved):

__mocks__/svgMock.js
```diff
@@ -0,0 +1,11 @@
...
+
+const svgMock = (...props) => {
+  const svg = React.createElement("svg", {
+    fill: "currentColor",
+    xmlns: "http://www.w3.org/2000/svg",
+    ...props[0],
+  });
+  return svg;
+};
+export default svgMock;
```
devharris7, line 11: eof

PR #18, opened 2022-05-17
fix: removed duplicate chakra provider import as this is already impo…

tatoMa:
fix: removed duplicate chakra provider import as this is already imported from styles/chakra file


devharris7 (approved):
LGTM

GitHubEason (approved):
LGTM

PR #19, opened 2022-05-17
feat: cc-0007 Show PriceBar on Blank Page, Add open close button

qiying417:
Add open close button to the pricebar to match the UI design.

image


JasonCaoAU (approved):
LGTM

src/components/PriceBar/index.tsx
```diff
@@ -43,6 +45,9 @@ const PriceBar: React.FC = () => {
   return (
     <>
       <Container>
+        <OpenCloseButton>
+          <AiOutlineDown />
+        </OpenCloseButton>
```
tatoMa, line 50: Is this for card cc007 or cc008? If for cc008, the button should open and close the price bar as I read from our notion board.
The solution can be applied by using a state conditionally render the price bar.

tatoMa (commented):
Also maybe move the testing files into the components folder.

src/components/PriceBar/index.tsx
qiying417: This is for cd007 AC1, "I can see the price bar with “v” on the top." Just for UI purpose.
The open close button feature which open/close the price bar will be implemented in cc008.

tatoMa (approved):

src/components/PriceBar/index.tsx
```diff
@@ -62,7 +67,7 @@ const PriceBar: React.FC = () => {
             <FormLabel htmlFor="ruler-on" mb={0}>
               Ruler On
             </FormLabel>
-            <Switch id="ruler-on" />
+            <Switch id="ruler-on" colorScheme={"blue"}/>
```
devharris7, line 70: remove `{}`

devharris7 (commented):
add a single

PR #21, opened 2022-05-18
fix: cc-0006 remove animation

stkwn:
The fix/cc-0006-see-sidebar-remove-animation is ready to be merged into the main branch.

We fix the bug that animation occurs when refreshing the page. We remove the animation.

FIX006


JasonCaoAU (approved):
LGTM

GitHubEason (approved):
LGTM

PR #23, opened 2022-05-19
Feature/cc 0015 see footer of editor page

Jzeva:
We built the footer bar under the price bar which contains zoon in, zoom out buttons, ruler switch and help icon.

UI:
image


tatoMa (changes requested):
When the right bottom switch is checked, the background color changes to the same as the thumb color.

image

Also seems the label text cannot trigger to switch this button on and off.

src/__tests__/components/EditorFooter/index.test.tsx
```diff
@@ -0,0 +1,29 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import EditorFooter from "../../../components/EditorFooter";
```
Fuchih, line 2: I suggest you can use "@/" instead of "../../../", it might make path more clear.

JasonCaoAU (approved):
LGTM

src/components/EditorFooter/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+                bgColor: "brand.primary",
+              },
+            }}
+            defaultChecked
+            onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+              setRulerLabel(e.target.checked ? "Ruler on" : "Ruler off")
+            }
+            data-testid="switch-btn"
+          />
+        </FormControl>
```
tatoMa, line 61: The value of id inside of the Switch component should be matched with the value of htmlFor in FormLabel. The reason why doing that is when these 2 values matches, you can click the label(the text) to change on and off of the Switch.

tatoMa (changes requested):

src/components/EditorFooter/index.tsx
EmilyGong111: Thanks for your request. Switch id has changed to be matched with the value of htmlFor in FormLabel. Please have a check. And appreciate your explanation on how  they work together.

tatoMa (approved):
LGTM

PR #27, opened 2022-05-21
fix: transition of pricerbar overlaps with footer in wrong zindex order

tatoMa:
This is for fixing the transition of price bar overlaps with drawer in wrong zindex order.
image


Fuchih (approved):

Jzeva (approved):
Tao God

PR #31, opened 2022-05-23
feat: cc-0019 add check communication btw fe&be

stevenliu986:


src/store/reducer/counterSlice.ts
```diff
@@ -9,6 +10,11 @@ const initialState: CounterState = {
   value: 0,
 };
 
+export const getTestData = createAsyncThunk("test commu btw fe & be", async (state) => {
+  const response = await axios.get("http://localhost:8080/api/v1/test");
```
tatoMa: Becare of the port number if there is no .env file inside of the backend repo.

src/pages/index.tsx
```diff
@@ -7,14 +7,17 @@ import {
   clear,
   incrementByAmount,
   changeCounter,
+  getTestData,
 } from "../store/reducer/counterSlice";
 import HeaderLayout from "../layouts/HeaderLayout";
 import { Button, Box, Flex, Heading, Text } from "@chakra-ui/react";
 import ChangeCourtSize from "@/components/ChangeCourtSize";
+import { AppDispatch } from "@/store";
```
tatoMa, line 15: There is an import store inside of the _app.js file already. That should make the redux store global on all pages and components.

tatoMa (commented):
May need to modify these codes a little. You have done pretty well overall.

src/store/reducer/counterSlice.ts
```diff
@@ -1,4 +1,5 @@
-import { createSlice, PayloadAction } from "@reduxjs/toolkit";
+import { createAsyncThunk, createSlice, PayloadAction } from "@reduxjs/toolkit";
+import axios from "axios";
```
devharris7, line 2: why not use axios from `utils/axios.js`?

src/pages/index.tsx
```diff
@@ -1,20 +1,21 @@
...
+  getTestData,
 } from "../store/reducer/counterSlice";
 import HeaderLayout from "../layouts/HeaderLayout";
 import { Button, Box, Flex, Heading, Text } from "@chakra-ui/react";
 import ChangeCourtSize from "@/components/ChangeCourtSize";
 
 const Home: NextPage = () => {
-  const count = useSelector(changeCounter);
-  const dispatch = useDispatch();
+  const count = useAppSelector(changeCounter);
```
tatoMa, line 17: It seems like this redux selector is for getting value from the store so the name of the selector might need to change.

tatoMa (approved):
I read through your code all. High-quality codes I can see. Only a very small issue and please ensure to provide .env file for our members.
LGTM

Fuchih (approved):

PR #33, opened 2022-05-25
Feature/cc 0020 a user can view blueprint

stkwn:
CC-0020 A user can view the blueprint page

When click on the blueprint on the sidebar, can see the blueprint page.
When hovering over the blueprint, can see the scrollbar as well
The user can scroll down and up the page and see the 6 images.
CC-0020


src/pages/style.css
```diff
@@ -0,0 +1,19 @@
...
+.scrollbox::-webkit-scrollbar {
+        width: 13px;
+        height: 15px;
+    }
+.scrollbox::-webkit-scrollbar-thumb  
+    {  
+    height: 15px;
+    border-radius: 5px;  
+    background-color: #ABABAD;  
+    }
```
tatoMa, line 19: I don't think using a global CSS is a good way to approach this request. Maybe try to define styles inside of variants in the Chakra styles folder as others do?

src/styles/components/blueprintsStyles.ts
```diff
@@ -0,0 +1,25 @@
...
+  // styles for different visual variants ("outline", "solid")
+  variants: {},
+  // default values for `size` and `variant`
+  defaultProps: {
+    size: '',
+    variant: '',
+  },
+}
+
+*/
```
tatoMa, line 25: Why are all these variants empty?

tatoMa (changes requested):
I have used Chrome and Safari to run this branch in dev mode. All of them are not work while using the mouse scroll wheel and the touchpad.

src/pages/style.css
stkwn: It seems that  Chakra  doesn't support some props like  the -webkit-scrollbar-thumb, might find a way to update it later.

src/styles/components/blueprintsStyles.ts
stkwn: just remove them.

src/components/EditorSideBar/EditorSideBarContent.tsx
```diff
@@ -30,7 +38,7 @@ const SideBarContainer = (props: Props) => {
         w="56px"
         bg="background.secondary"
         clipPath="polygon(0% 0%, 100% 8%, 100% 92%, 0% 100%)"
-        z-index="1"
+        zIndex="-1"
```
Jzeva, line 41: What effect did you want to achieve by adding z-index here

tatoMa (approved):
LGTM

Jzeva (approved):

src/components/EditorSideBar/EditorSideBarContent.tsx
stkwn: It's a sidebar close button, I need to put it -1, otherwise, it will be on top of the normal sidebar content box.

tatoMa (approved):

tatoMa (approved):

PR #34, opened 2022-05-25
Feature/cc 0018 a user can see the 404 page

JasonCaoAU:
feature/cc-0018-user-can-see-the-404page is ready to be reviewed and moved into main branch. 
The file 404.tsx can be found in the pages directory. 
So, user on the 404page can see the number 404 with the test" Oops! Something went wrong", and user can click the "Take me back to the homepage" button to be back to the homepage.
image


src/test-utils/createMockRouter.ts
```diff
@@ -0,0 +1,29 @@
...
+    },
+    isFallback: false,
+    isLocaleDomain: false,
+    isReady: true,
+    defaultLocale: 'en',
+    domainLocales: [],
+    isPreview: false,
+    ...router,
+  };
+}
```
tatoMa, line 29: Consider moving this file into _test_/utils.tsx

src/layouts/index.tsx
```diff
@@ -1,14 +1,24 @@
...
 
 const Layout: React.FC<{ children: ReactNode }> = ({ children }) => {
+  const router = useRouter();
+  if (router.pathname === PAGE_NOT_FOUND) {
+    return (
+      <Box as="main">
+        <NavigationBar />
+        {children}
+      </Box>
+    );
```
tatoMa, line 20: I think you also need a Header(NextHeadSeo or Header for seo) component here

src/pages/404.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+export default function Custom404() {
+  return (
+    <>
+      <NextHeadSeo
+        title={`404 Error page - Courtcanva`}
+        description={`404 Error page on CourtCanva`}
+        canonical={`http://www.courtcanva.com/404`}
+      />
+      <Flex direction="column" justify="center" align="center" h="calc(100vh - 106px)">
```
tatoMa, line 15: Why 100vh - 106px? The navbar height is 72px I believe.

src/pages/404.tsx
```diff
@@ -0,0 +1,31 @@
+import Link from "next/link";
+import { Flex, Button, Text } from "@chakra-ui/react";
+import UnSmile from "@/assets/svg/404.svg";
+import HOME_PAGE_LINK from "../constants";
```
tatoMa, line 4: The home page link may not be https://courtcanva.com/. It may go to the UAT version home page.
Using a variable in .env file to define this maybe a better approach.

tatoMa (changes requested):

src/__tests__/layouts/Layout.test.tsx
```diff
@@ -1,13 +1,35 @@
...
-      <Layout>
-        <Home />
-      </Layout>
+  it("should render the navBar and 404", () => {
+    render(
+      <RouterContext.Provider value={createMockRouter({ pathname: "/404" })}>
+        <Layout>
+          <Custom404 />
+        </Layout>
+      </RouterContext.Provider>
```
Jzeva, line 14: Persoanally, I think this description is not correct. I think it should focus on the layout itself. Also, I think  custom404 should not be here becasue we are writing unit test for "Layout"

src/test-utils/createMockRouter.ts
Jzeva: Agree

src/pages/404.tsx
JasonCaoAU: Yes you are right, i have fixed it. As we considered the footer hight before.

src/pages/404.tsx
JasonCaoAU: it has been fixed , thanks

src/pages/404.tsx
```diff
@@ -0,0 +1,31 @@
+import Link from "next/link";
+import { Flex, Button, Text } from "@chakra-ui/react";
+import UnSmile from "@/assets/svg/404.svg";
+import NextHeadSeo from "next-head-seo";
+process.env.NEXT_PUBLIC_UAT_URL;
```
tatoMa, line 5: Delete this line.

JasonCaoAU: It has beed deleted.

PR #39, opened 2022-06-02
Feature/cc 0021 users can view three point area

Jzeva:
UI:
image

We initiated the canva and drew the three point line area on the canva.


package.json
```diff
@@ -79,6 +81,7 @@
     "jest": "^27.5.1",
     "lint-staged": "^12.3.7",
     "prettier": "^2.6.2",
+    "react-test-renderer": "^17.0.2",
```
devharris7, line 84: do we need this dep?

tatoMa, line 84: Seems this lib is not used for now?

src/__tests__/components/BasketballCourt/helper.tsx
```diff
@@ -0,0 +1,36 @@
...
+import React from "react";
+import ReactDOM from "react-dom";
+
+type Component = React.ReactNode;
+interface AppProps {
+  onUpdate: (prop: unknown) => void;
+  children: Component;
+}
+
+const render = async (component: Component) => {
```
devharris7, line 11: `import {ReactNode} from 'react`
`component: ReactNode`

seems you don't need to reassign a type name to it

src/__tests__/redux/courtSlices/threePointLineSlice.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+} from "@/store/reducer/courtSlices/threePointLineSlice";
+import store from "@/store/index";
+
+it("should return the initial state", () => {
+  const state = store.getState().threePointLine;
+  expect(state).toEqual(initialState);
+});
+
+it("should handle coordinates", () => {
+  // temporary solution, should be changed later
```
devharris7, line 14: Add `TODO: `

src/layouts/index.tsx
```diff
@@ -31,8 +31,8 @@ const Layout: React.FC<{ children: ReactNode }> = ({ children }) => {
         <EditorSideBar />
         {children}
         <TopBar />
-        <Footer />
       </Box>
+      {/* <Footer /> */}
```
tatoMa, line 35: If this footer is not used anymore and no UI need to use it, can simply delete it with using and import.

src/components/BasketballCourt/ThreePointArea.tsx
```diff
@@ -0,0 +1,54 @@
...
+          controlPointThreeY,
+          controlPointFourX,
+          controlPointFourY,
+          threePointLineRadius
+        );
+        // Continue with vertical line which makes it a close shape
+        context.lineTo(30, 266);
+
+        context.closePath();
+        // (!) Konva specific method, it is very important
```
devharris7, line 44: it's better to leave a doc link for this method, rather than say it's important

src/pages/style.css
```diff
@@ -1,19 +1,25 @@
-.scrollbox{
-    overflow: hidden;
+html,
+body {
+  background-color: #EBECF0;
```
devharris7, line 3: do you want to keep color somewhere together

src/pages/index.tsx
```diff
@@ -1,63 +1,15 @@
...
-  incrementByAmount,
-  changeCounter,
-  getTestData,
-} from "../store/reducer/counterSlice";
 import HeaderLayout from "../layouts/HeaderLayout";
-import { Button, Box, Flex, Heading, Text } from "@chakra-ui/react";
-import ChangeCourtSize from "@/components/ChangeCourtSize";
+import dynamic from "next/dynamic";
+
+const BasketballCourt = dynamic(() => import("@/components/BasketballCourt"), { ssr: false });
```
tatoMa, line 5: Please explain why needs lazy loading on this component?

devharris7 (approved):
Code looks good, but requires manual testing by some other devs.
Please ask someone to pull your code and test it locally, then add his review comments to your PR

devharris7:
Add a label `peer reviewed`, you can add it when someone help you test it

src/components/BasketballCourt/ThreePointArea.tsx
```diff
@@ -0,0 +1,54 @@
...
+        context.fillStrokeShape(shape);
+      }}
+      fill={color}
+      stroke="white"
```
tatoMa, line 48: Why only color is a prop with default value but stroke color and width are not?

package.json
Fuchih: forgot to delete it sorry about that

Fuchih: forgot to delete it sorry about that

src/pages/index.tsx
Jzeva: We treid to remove this lazy loading, but we found that it would throw an exception. The only way to fix it is to import it dynamically. Do you have any suggestions about that?
https://github.com/konvajs/react-konva/issues/588

tatoMa: I think you are right. I haven't found a useful solution for ESM mode on Nextjs yet. The official document may lead to a closer solution but it not working itself.
https://nextjs.org/blog/next-11-1#es-modules-support

src/components/BasketballCourt/index.tsx
```diff
@@ -0,0 +1,42 @@
+import { Stage, Layer } from "react-konva";
+import { Flex } from "@chakra-ui/react";
+import ThreePointArea from "./ThreePointArea";
+import { ReactReduxContext, Provider } from "react-redux";
+import { calcAxisDelta } from "framer-motion/types/projection/geometry/delta-calc";
```
tatoMa, line 5: Not used anymore?

tatoMa (approved):
I haven't found many helpful solutions for fixing the EMS/ CommonJS problem when this component is not dynamically imported into the index page. So I would say keep it in this way for now. Maybe later we can find a solution for that.

src/components/BasketballCourt/index.tsx
Jzeva: Thanks for reminding. I have deleted the unused import.

PR #47, opened 2022-06-08
Feature/cc 0025 a user can view the court area

EmilyGong111:
cc-0025 a user can view the court area is ready to be review. Thanks!


tatoMa (approved):
LGTM

PR #48, opened 2022-06-08
feat: cc-0022 first version & README

Beryl-Zhou:
keyArea
1. keyArea first version
2. ReadMe update

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,11 @@ You can start editing the page by modifying `src/pages/index.tsx`. The page auto
 
 The `src/pages/api` directory is mapped to `/api/*`. Files in this directory are treated as [API routes](https://nextjs.org/docs/api-routes/introduction) instead of React pages.
 
+### Prepare First Commit
+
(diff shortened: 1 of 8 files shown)


tatoMa (approved):
LGTM

README.md
```diff
@@ -26,6 +26,11 @@ You can start editing the page by modifying `src/pages/index.tsx`. The page auto
 
 The `src/pages/api` directory is mapped to `/api/*`. Files in this directory are treated as [API routes](https://nextjs.org/docs/api-routes/introduction) instead of React pages.
 
+### Prepare First Commit
+
+Add a file named .env at the root directory. Copy the JS from ticket [Frontend ENV](https://www.notion.so/CourtCanva-Boards-9a664c4a1114424080ed647b31873151?p=5c24cef3686b4426a7fd012b52d49fcd) and paste it in .env.
+
```
kritomomo, line 32: We have .env.example

Jzeva (approved):

README.md
Beryl-Zhou: Just double-check the code. The .env.example file does not work in this circumstance. It is advised .env file is necessary for the first commit.

kritomomo: Actually this topic is not about the first commit, it is about setting environment variables in the local environment.
In addition, the role of setting .env.example is to facilitate us to copy the variable name directly from .env.example when creating a local .env, or we can directly use "cp .env.example .env" to create it directly

PR #50, opened 2022-06-09
Feature/cc-0024-a-user-can-view-the-border-area

Jzeva:
We drew the border of the court on the canva.
UI:
image


tatoMa (approved):
LGTM. However, since the border is drawn, the whole court seems a lot smaller than the frame (canvas) we have now. May need to discuss it with group members about it.

Jzeva:
>

EmilyGong111 (approved):
LGTM

PR #55, opened 2022-06-12
Feature/cc 0042 a user can log in sign up for my account

jennawan:
login UI
login UI modified.
cc 0042 is ready for code review.


src/styles/theme.ts
```diff
@@ -27,18 +28,23 @@ const colors = {
...
-    secondary : "#4f5757",
-    third:"#EDF2F7"
+    secondary: "#4f5757",
+    third: "#EDF2F7",
   },
   fontcolor: {
     primary: "#C8C5C5",
-    secondary:"#1A202C",
+    secondary: "#1A202C",
+    third: "#4285f4",
```
Jzeva, line 37: Do you think there should be 'tertiary' instead of 'third'?

jennawan: Sure, followed "background.third", didn't think through it. Will change.

package.json
```diff
@@ -37,6 +37,8 @@
     "@chakra-ui/react": "^1.8.8",
     "@emotion/react": "^11",
     "@emotion/styled": "^11",
+    "@fontsource/inter": "^4.5.11",
+    "@fontsource/lobster-two": "^4.5.8",
```
tatoMa, line 41: Give me a reason why using font lib instead of using the font from google font on CSS by using font-face.

src/__tests__/components/Login/index.test.tsx
```diff
@@ -0,0 +1,19 @@
+import LoginModalContent from "@/components/Login";
+import { render, screen } from "@testing-library/react";
+
+describe("Login", () => {
+  test("Each elements in the Login display the correct text", () => {
+    render(<LoginModalContent isOpen={true} onClose={() => void {}} />);
```
kritomomo, line 6: You can use jest mock functions, check this [link ](https://jestjs.io/docs/jest-object#jestfnimplementation)

src/components/Login/index.tsx
```diff
@@ -31,40 +38,44 @@ function LoginModalContent(props: Props) {
       <ModalOverlay />
       <ModalContent>
         <ModalHeader>
-          <Flex flexDir="column" mt="40px" mb="20px">
-            <Text fontSize="xl" fontWeight="bold">
-              Log in or sign up
+          <Flex flexDir="column" alignItems="center" mt="40px">
+            <Text sx={logoStyles} fontSize="2xl" mb="20px" role="logo">
```
kritomomo, line 42: Can we use variant here？
image
https://chakra-ui.com/docs/components/typography/text

package.json
jennawan: Sorry, I don't have a particularly good reason. Install with NPM is the first option provided by Chakra UI and also the first option that worked for me. Thinking through now, using font-face would be the better option. Will give font-face a go.

src/components/Login/index.tsx
```diff
@@ -19,6 +25,7 @@ interface Props {
 
 function LoginModalContent(props: Props) {
   const initialRef = React.useRef(null);
+  const logoStyles = useStyleConfig("Logo");
 
   return (
     <Modal
```
kritomomo, line 31: When I open the modal, the close button will auto focus
image

tatoMa: Try to find the way of using "initialFocusRef" in the Modal component on Chakra Doc can fix this.

Normally people would not like to focus on the close button after opening a modal because users may just hit the enter key and assume this is for the first option(Continue with Google in our case)

EmilyGong111:
Screen Shot 2022-06-12 at 9 34 11 pm
Screen Shot 2022-06-12 at 9 34 21 pm

The sign in component box may need more padding, it looks a little bit too full to me. The contrast of font size and width between different element  may need a little bit modification.

src/components/Login/index.tsx
Kornzk: You can refer to this part we did before.

Kornzk: You can refer to this part we did before.cc-0035

package.json
jennawan: font packages have been removed as per discussed.

src/__tests__/components/Login/index.test.tsx
jennawan: Tests have been coded without using jest.fn() as per discussed.

src/components/Login/index.tsx
jennawan: logoStyles was not used as logo text was replaced with svg.

src/components/Login/index.tsx
jennawan: Fixed. Thank you all for your suggestion and advice.

tatoMa (approved):
LGTM now. Nice work.

tatoMa:
> Use your email or Google account to continue with CourtCanva! Use your Google account or email to continue with CourtCanva! In which order is better? The latter corresponds to the button order.

I would say that google login is more important than personal email at this stage. As our project is just a brand new one, using google OAuth for sign-in and sign-up is easier for users to do, and I think it's not going to have much difference on our end cos we store whatever users give us, Gmail or any email.

All just my opinion.

Jzeva (approved):

PR #61, opened 2022-06-15
Feature/cc 0027 a user can view pro full court

Beryl-Zhou:


EmilyGong111 (approved):
Great job! Really learned a lot. Thanks!

tatoMa (approved):
Brilliant job! You have done so well.

Just some thought. Maybe added a few comments for group members to easier understand what you have done, especially for those complicated formulas and functions.

If there are any methods, formulas, or functions you wrote twice, maybe better to extract them into utility files.

Overall, great work.

src/components/ProFullCourt/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+                <ArrowLine courtRatio={courtRatio} arrowXEndLength={courtAreaXLength} />
+                <Group>
+                  <BorderDimensionLine courtRatio={courtRatio} />
+                  <CourtArea courtRatio={courtRatio} color={"#B61313"} />
+                  <ThreePointArea courtRatio={courtRatio} color={"#72818B"} />
+                  <KeyArea courtRatio={courtRatio} color={"#2C4E8A"} />
+                  <CircleArea courtRatio={courtRatio} color={"#606F14"} />
+                  <TopKeyArea courtRatio={courtRatio} color={"#B61313"} />
+                </Group>
+                <Group scaleX={-1} x={initPointX * 2 + courtAreaXLength * 2 * courtRatio}>
```
tatoMa, line 61: In this part, if I'm not wrong, you pass the formula result to the whole group and combine using scaleX to flip the half of the court horizontally?

Do you think is there any easy-to-understand way of doing it?
It is just a though, not a change request.

Beryl-Zhou: Sure. now I am still looking for a better way to simplify the code. A fix branch may be added once an easy way has been found to achieve the goal. Cheers~~~

PR #63, opened 2022-06-16
fix: cc-0027 delete easy-in-out effect

Beryl-Zhou:


src/layouts/HeaderLayout.tsx
```diff
@@ -18,11 +18,6 @@ const HeaderLayout: React.FC<HeaderLayoutProps> = ({ children, title }) => {
   const headName = `${title} | CourtCanva`;
   return (
     <motion.article
```
tatoMa, line 20: Yes, this is the correct place for removing the transition of our pages but do you think we still need an empty motion.article tag?

src/layouts/HeaderLayout.tsx
```diff
@@ -17,14 +17,6 @@ const variants = {
 const HeaderLayout: React.FC<HeaderLayoutProps> = ({ children, title }) => {
   const headName = `${title} | CourtCanva`;
   return (
-    <motion.article
-      initial="hidden"
-      animate="enter"
-      exit="exit"
-      variants={variants}
```
kritomomo, line 24: image
Looks like we don't need variants now

kritomomo (approved):
LGTM, Remember squash your commit message before merge

tatoMa (approved):
LGTM now.

PR #64, opened 2022-06-17
Feature/cc 0048 a user can view the price bar

jennawan:
price bar
Please help review the code for cc-0048 a user can view the price bar. Cheers!


tatoMa (commented):
Why the colors of some areas in canvas are different as our UI design?
image

tatoMa (commented):
Think about future features while we need to change the individual color of tiles, you have 2 areas bound in one color variable. Maybe hard to change it later to different colors?
image

tatoMa (approved):

Beryl-Zhou (approved):

PR #66, opened 2022-06-17
Refactor/check connection febe

GitHubEason:
Move health check to a new page
Screen Shot 2022-06-18 at 12 09 16 am


jennawan (approved):
Works fine on my side, nice job!

src/pages/health.tsx
```diff
@@ -0,0 +1,31 @@
+import { useState, useEffect } from "react";
+import { Flex, Button } from "@chakra-ui/react";
+import { createAsyncThunk, createSlice, PayloadAction } from "@reduxjs/toolkit";
+import { api } from "../utils/axios";
+import { useStoreDispatch } from "../store/hooks";
+
+const healthCheck = createAsyncThunk("Health Check", async () => {
+  api(process.env.NEXT_PUBLIC_HEALTH_API as string, { method: "get" })
+    .then((response) => {
+      console.log(response);
```
jennawan, line 10: Considering removing the console.log?

jennawan (approved):

.env.example
```diff
@@ -1,4 +1,3 @@
-NEXT_PUBLIC_UAT_URL=
-NEXT_PUBLIC_TEST_API=
+NEXT_PUBLIC_DESIGN_URL=
```
tatoMa, line 1: Why it is called DESIGN URL? I thought it's for the home page(marketing page) URL.

GitHubEason: I think  this URL for the 404-page redirects to the design page.

GitHubEason: Thanks for reminding me

tatoMa (commented):
No test file for this page yet so it cannot pass the coverage request.

Jzeva:
If you are sure that you don't need to add tests or exclude it in the test config then it looks OK to me.

kritomomo:
I don't think the front end needs this health check

PR #72, opened 2022-06-18
fix: 404 page redirection url name

GitHubEason:


Jzeva:
Personally I cannot see any necessity for this change, because this url still points to our UAT environment no matter you rename it or not. (For me this name is even more confusing than before) Of course if you find any naming convention for this please put the link here, I will learn it more carefully.

ki-ki-316 (approved):
Looks good

Jzeva (approved):

PR #73, opened 2022-06-18
Feature/cc 0055 remove court thumbnail from top bar

jennawan:
UI of top bar updated. Court and court thumbnail and bottom shadow removed from top bar.
image


tatoMa (approved):
Nice work. It's so nice that you change all prop variable names from abbreviation back to full name.

Fuchih (approved):

PR #80, opened 2022-06-23
feat: CC-0057 A user can use blueprint function

Kornzk:
截屏2022-06-23 下午9 07 47
截屏2022-06-23 下午9 07 30

When I hover over the picture of court sizes in the blueprint 
I can see the picture with a green frame

When I click the picture
I can see other pictures except for the one I clicked become darkened.


tatoMa (commented):
Do you think it's better from a UX point of view that after you click and choose one size then the menu closes automatically?
And if you have chosen one size (including the default first one) it could be highlighted on the list so that users know which one they have selected now?

EmilyGong111 (approved):
LGTM

jennawan (approved):

PR #81, opened 2022-06-24
Feature/cc 0006 automatic calculation of the tile number

jennawan:
test not done yet.


src/components/ProFullCourt/index.tsx
```diff
@@ -30,6 +34,32 @@ const ProFullCourt = () => {
...
   const [size, setSize] = useState({ width: window.innerWidth, height: window.innerHeight });
 
+  const courtAndTileInfo = {
+    beginPointX: (stageMargin - borderLength) * court.courtRatio,
+    beginPointY: (stageMargin - borderLength) * court.courtRatio,
+    endPointX: (stageMargin + courtAreaXLength + borderLength) * court.courtRatio,
+    endPointY: (stageMargin + courtAreaYLength + borderLength) * court.courtRatio,
+    // TO CHANGE LATER: tile size will be passed in instead of hard coding
+    tileSize: 300 * court.courtRatio,
+  };
```
tatoMa, line 44: Do you think other courts also need this formula for the same calculation? If yes, maybe extract it into a common util file for use on all the other courses.

src/components/ProFullCourt/index.tsx
```diff
@@ -30,6 +34,32 @@ const ProFullCourt = () => {
...
+  const debouncedCalculation = useCallback(
+    debounce(() => {
+      canvas = canvasRef.current as unknown as HTMLCanvasElement;
+      if (canvas) {
+        ctx = canvas.getContext("2d");
+        const tileNumResult = tileNumberCalculator(ctx, courtAndTileInfo);
+        // To Delete later, console for preview only
+        console.log(tileNumResult);
+      }
+    }, 500),
```
tatoMa, line 59: This can be a common utility as well I think.

src/components/ProFullCourt/index.tsx
```diff
@@ -30,6 +34,32 @@ const ProFullCourt = () => {
...
+    []
+  );
+
   useLayoutEffect(() => {
```
tatoMa, line 63: This is a great way to use useLayoutEffect. Nice work!

tatoMa (commented):
I kind of get what you are trying to do in the tile number calculator. But my question is that there is only a giant method inside of it. Do you think you also need to extract some of the parts to make it into a few methods that are easier to use and test? (Just my thought)

EmilyGong111 (approved):
Nice work.
Just remind.
When you extract the logic in ProFullCourt to a common util, make sure if borderLength is undefined would not affect the result.
Because currently other courts does not have border.

src/components/ProFullCourt/index.tsx
jennawan: Thank you for this advice. Other courts would need calculation but with different formulas, as this ProFullCourt has a border that other courts do not have. In a future card, the calculation for other courts (if same), will be extracted to a util file.

src/components/ProFullCourt/index.tsx
jennawan: Thank you! Done.

tatoMa (commented):
Nice work of refactoring the calculator utils.  😀

Beryl-Zhou (approved):

src/utils/tileNumberCalculator.ts
```diff
@@ -0,0 +1,83 @@
...
+
+// set the color that takes most pixels as the tile color
+// by finding the most repeated element in pixelColorInTile array
+const determineTileColor = (arr: Array<string>) => {
+  const hash: hashObject = {};
+  let maxNum = 0;
+  let mostEle = null;
+  for (let i = 0 as number; i < arr.length; i++) {
+    // ignore white, black and blank color
+    arr[i] === "0,0,0,0" || arr[i] === "0,0,0,255" || arr[i] === "255,255,255,255"
```
devharris7, line 34: ["0,0,0,0", '0,0,0,255', '255,255,255,255'].includes(arr[i])

devharris7: if ()

src/utils/tileNumberCalculator.ts
```diff
@@ -0,0 +1,83 @@
...
+};
+
+export const tileNumberCalculator = (
+  ctx: CanvasRenderingContext2D | null,
+  courtAndTileInfo: IcourtAndTileInfo
+) => {
+  const { beginPointX, beginPointY, endPointX, endPointY, tileSize } = courtAndTileInfo;
+  const rgbaResult: hashObject = {};
+  // scan horizontally
+  for (let x = beginPointX; x < endPointX; x += tileSize) {
```
devharris7, line 55: 双指针？

devharris7: @chui-mel

PR #88, opened 2022-06-27
Feature/cc 0065 correct court size displace error

stkwn:
CC-0065 Correct court size displace error 
the original one is 
0065-o

update it to

### DIFF
--- a/src/store/reducer/courtNameSlice.ts
+++ b/src/store/reducer/courtNameSlice.ts
@@ -5,7 +5,7 @@ export interface CourtNameState {
 }
 
 export const initialState: CourtNameState = {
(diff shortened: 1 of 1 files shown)


tatoMa (commented):
This is only for the init redux value. What about the list of our mock data? (inside of CourtList file)

tatoMa (approved):

Kornzk (approved):

EmilyGong111 (changes requested):
You may need to do a global check. Because the data is base on mocked data and is also used in other documents.

EmilyGong111 (approved):

PR #93, opened 2022-06-29
feat: cc-0072 combine the same area and prevent default color change …

EmilyGong111:
Same court area will change color together.


tatoMa (approved):
ALl LGTM

Fuchih (approved):

PR #94, opened 2022-06-29
Feature/cc 0073 connect court color change with tiles slice

EmilyGong111:
connect court color with price bar tilecolor board
Screen Shot 2022-06-29 at 4 21 27 pm

### DIFF
--- a/src/store/reducer/tileSlice.ts
+++ b/src/store/reducer/tileSlice.ts
@@ -2,34 +2,39 @@ import { createSlice, PayloadAction } from "@reduxjs/toolkit";
 import { RootState } from "..";
 
 export interface tileState {
-  location: Array<string>;
+  location: string;
   color: string;
   quantity: number;
 }
 
 export const initialState: tileState[] = [
   {
-    location: ["threePoint"],
+    location: "threePoint",
     color: "#72818B",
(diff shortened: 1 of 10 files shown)


jennawan (approved):
LGTM😄

src/store/reducer/tileSlice.ts
```diff
@@ -39,8 +44,12 @@ export const tileSlice = createSlice({
...
   initialState,
   reducers: {
-    changeTileColor: (state, action: PayloadAction<string>) => {
-      // TODO COLOR CHANGE
+    /* istanbul ignore next */
+    changeTileColor: (state, action: PayloadAction<any>) => {
+      const selectedLocation = state.findIndex(
+        (object) => object.location === action.payload.location
+      );
+      state[selectedLocation].color = action.payload.selectedColor;
```
tatoMa, line 52: Please remember, redux tool kit uses immer.js which is create an immutable value based on the current values under the hood. If there is no immer, then you cant modify the state, you have to create a new one instead of the old one. 
(It is not a change requirement if you `are` interested in the theory, it you can have a look at immer.js)

tatoMa (approved):
LGTM

PR #96, opened 2022-06-29
Feature/cc 0071 blueprint highlights bug

Kornzk:
fix blueprint highlights bug


Fuchih (approved):

src/components/EditorSideBar/Blueprints.tsx
```diff
@@ -35,7 +35,7 @@ const Blueprints: React.FC = () => {
             onClick={() => handleCourtSelecting(courtSizeName, courtSizeDetails, img)}
             data-testid={img}
             _hover={{ border: "4px solid #40B484" }}
-            opacity={activateCourt === img ? "1" : "0.4"}
+            opacity={!activateCourt || activateCourt === img ? "1" : "0.4"}
```
tatoMa, line 38: Even though it works, however it's not a good idea to put complicated conditions inline in the component props. Better to extract it and make a method that returns the result for it.

tatoMa (approved):

EmilyGong111 (approved):

PR #100, opened 2022-07-01
Feature/cc 0052 sum up tile quantity and update price bar

jennawan:
image
image
image
Price bar estimated tiles are able to change with the color change of court, and the tile number is summed up based on the result from tileNumberCalculator.


src/components/ProFullCourt/index.tsx
```diff
@@ -14,6 +14,9 @@ import { useStoreSelector } from "@/store/hooks";
 import DashedLine from "../BasketballCourt/DashedLine";
 import BorderDimension from "../BasketballCourt/BorderDimensionLine";
 import { debouncedCalculation } from "@/utils/tileNumberCalculator";
+import { useDispatch } from "react-redux";
+import { changeTileQuantity } from "@/store/reducer/tileSlice";
+import { clearTimeout, setTimeout } from "timers";
```
tatoMa, line 19: If I'm not wrong. This timers lib is from Node. When you use it on frontend, especially using Nextjs, it will work. But I don't see any difference between timers and web API clearTimeout/setTimeout.
Or if you have any reason using this lib?

jennawan: Thanks! Imported timer lib by mistake. Removed.

src/components/ProFullCourt/index.tsx
```diff
@@ -58,7 +62,14 @@ const ProFullCourt = () => {
...
 
-  tileCalculation(canvasRef, courtAndTileInfo);
+  const dispatch = useDispatch();
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      const tileNumberResult = tileCalculation(canvasRef, courtAndTileInfo);
+      dispatch(changeTileQuantity(tileNumberResult));
+    }, 100);
+    return () => clearTimeout(timer);
+  }, [tileColorState]);
```
tatoMa, line 72: When tileColorState changes, you set up a timeout and calculate the tile number result and dispatch redux function after 100ms, why do this?

tatoMa, line 72: I dont get why you run this calculation and dispatch after 100ms. Please explain.

jennawan: If I remove setTimeout (you can try to comment them out),  and run this calculation and dispatch at the same time tileColorState changed, the data that are being calculated and dispatched are still the state before the color change. There will be a bug that result shown in price bar will be the previous result. setTimeout is the solution for this bug.

tatoMa: Yes, you are right. Seems like this calculation needs time to send the result back so the whole calculation and dispatch need to be async instead of sync. Even using "0" as the second parameter in setTimeout will do the trick as well. I will look deep into this function tonight or tomorrow to see if there is any proper way to fix this issue. Otherwise, nice work. 👍

src/components/BasketballCourt/CourtArea.tsx
```diff
@@ -15,7 +15,7 @@ const CourtArea: React.FC<CourtAreaProps> = ({ courtWidth, startPoint }) => {
 
   const selectedColor = useStoreSelector((state) => state.courtColor.selectedColor);
   const courtAreaColor = useStoreSelector(
-    (state) => state.tile.find((tile) => tile.location.includes("courtArea"))?.color
+    (state) => state.tile.court?.find((tile) => tile.location.includes("courtArea"))?.color
```
kritomomo, line 18: I think you can extract this part of the logic into a reusable function, because I see you use the same method in other files, and this improves the readability of the code
image

jennawan: Thank you! Done.

EmilyGong111 (approved):

Jzeva (approved):

PR #104, opened 2022-07-02
feat: CC -0069 Add tile calculation to each court

Jzeva:
Apply calculation to all courts and extract get data function into utils
image
image


tatoMa (commented):
Seems like there is something wrong with a few of these components, or it is related to calculation.
image

Jzeva:
> Seems like there is something wrong with a few of these components, or it is related to calculation. image

Yes, it seems like there is something wrong with it. I will double check with Jenna and try to fix it.

jennawan (approved):
The bug seems fixed now. LGTM.

Beryl-Zhou (approved):

PR #106, opened 2022-07-05
feat: cc-0076 use RTK query get data from backend

Kornzk:
For example: when you click the sidebar,  you can get court data and store it in the redux.
截屏2022-07-05 上午9 18 23
截屏2022-07-05 上午9 18 51


stevenliu986 (approved):
LGTM

EmilyGong111 (approved):
LGTM

PR #108, opened 2022-07-05
Feature/cc 0016 user can download design

EmilyGong111:
Click the download btn on topbar, can download the designed court in PDF format.
Screen Shot 2022-07-05 at 9 18 08 pm


src/components/TopBar/index.tsx
```diff
@@ -13,15 +13,17 @@ import { Popover, PopoverTrigger, PopoverContent, PopoverBody } from "@chakra-ui
...
 
 import ColorBoard from "./ColorBoard";
+import DownloadSvg from "@/assets/svg/TopBarSvg/download.svg";
 import BinSvg from "@/assets/svg/TopBarSvg/bin.svg";
 import DocSvg from "@/assets/svg/TopBarSvg/document.svg";
 import PaintBucketSvg from "@/assets/svg/TopBarSvg/paintBucket.svg";
 import UploadSvg from "@/assets/svg/TopBarSvg/upload.svg";
 import { useDispatch } from "react-redux";
 import { changeSelectedColor } from "@/store/reducer/courtColorSlice";
+import { downloadToPDF } from "../../utils/printPDF";
```
tatoMa, line 23: Maybe try dynamic(lazy) load this util. If it can be lazy-loaded then we don't need to bundle this lib to our main js file after build.

stevenliu986 (approved):
LGTM

src/components/TopBar/index.tsx
tatoMa: Cos it has 2 libraries in it. Lazy loading in any form would be better than bundle them all into our main exported file.

src/components/NavBar/EditorDesignName.tsx
```diff
@@ -8,11 +9,13 @@ import {
   EditableInput,
 } from "@chakra-ui/react";
 import { useState } from "react";
-
+import _ from "lodash";
```
Fuchih, line 12: looks like you don't need to import lodash

src/components/TopBar/index.tsx
EmilyGong111: Got it. I tried dynamic loading to import the function, but it was not allow. Cos dynamic import is used to import component. I will try another way.

src/store/reducer/designNameSlice.ts
```diff
@@ -0,0 +1,23 @@
+import { createSlice, PayloadAction } from "@reduxjs/toolkit";
+
+export interface DesignNameState {
+  name: string;
+}
+
+export const initialState: DesignNameState = {
+  name: "Court Canva 1",
```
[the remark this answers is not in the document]
EmilyGong111: I checked it, it is same. Not sure how it happens. Could you try it again?

src/components/TopBar/index.tsx
jennawan: Done. Followed document from this link https://javascript.plainenglish.io/a-lazy-load-util-function-you-will-get-addicted-to-e682ab410ac8. using react/webpack approach.

PR #109, opened 2022-07-05
feat: cc-0066 change stroke color when hovering any part of court

Fuchih:
change stroke colour when hovering over any part of the court to know which area can be filled with colour.

1. three-point
image

2. court-area
image

3. top-key
image

4. key-area
image

5. circle-area
image


EmilyGong111:
Nice coding. But this feature seems conflict with tile calculation. Please check the picture below.
WeChat09d34681be0627e79c8b7520cd615133

tatoMa:
> Nice coding. But this feature seems conflict with tile calculation. Please check the picture below. WeChat09d34681be0627e79c8b7520cd615133

Seems this kind of implementation is not suitable for our project. The tile calculation needs to find the color on canvas so that any way of changing color would have a negative for the calculation.
Maybe consider using other ways. Such as CSS or adding something that would not affect the calculation.

Fuchih:
> > Nice coding. But this feature seems conflict with tile calculation. Please check the picture below. WeChat09d34681be0627e79c8b7520cd615133
> 
> Seems this kind of implementation is not suitable for our project. The tile calculation needs to find the color on canvas so that any way of changing color would have a negative for the calculation. Maybe consider using other ways. Such as CSS or adding something that would not affect the calculation.

just do some copy and pasting from tileSlice to see the result, still in progress and those issues will be fixed later. Thanks for review

Fuchih:
> Nice coding. But this feature seems conflict with tile calculation. Please check the picture below. WeChat09d34681be0627e79c8b7520cd615133

just do some copy and pasting from tileSlice to see the result, still in progress and those issues will be fixed later. Thanks for review

EmilyGong111 (approved):

PR #110, opened 2022-07-06
feat: cc-0081-renew-court-data-and-related-codes

stevenliu986:


src/components/ChangeCourtSize/CourtList.tsx
```diff
@@ -1,30 +1,36 @@
 const courtList = [
   {
+    courtId: "62c432cfb8a9c5f61f03831f",
     courtSizeName: "510 m² Pro Court",
     courtSizeDetails: "(17 m × 30 m)",
```
EmilyGong111, line 5: We may do not need courtSizeName and courtSize details here. Cos the name we can get from courtSizeSlice and the numbers are variables( border changes, the number changes).  Same as below.

EmilyGong111: Screen Shot 2022-07-06 at 4 35 50 pm
I understand the courtSizeName and courtSizeDetails are used in the string showing on the picture. We may change the logic, using concatenated string by adding the data variables with the rest characters to show the content above.

src/components/EditorSideBar/Blueprints.tsx
```diff
@@ -1,30 +1,54 @@
 import courtList from "../ChangeCourtSize/CourtList";
 import { Image, Box } from "@chakra-ui/react";
 import { useDispatch } from "react-redux";
-import { changeCourtName } from "@/store/reducer/courtNameSlice";
+import { changeCourtName, CourtNameState } from "@/store/reducer/courtNameSlice";
 import React, { useState } from "react";
-import { useGetCourtsQuery, useGetCourtQuery } from "@/pages/api/courtSizeApi";
+import { useGetCourtsQuery } from "@/pages/api/courtSizeApi";
```
tatoMa, line 6: Please move this file into redux/api

EmilyGong111 (approved):

Fuchih (approved):

PR #112, opened 2022-07-09
Feature/cc 0090 undo and redo function

jennawan:
image
Both buttons are disabled when page is opened.
image
Undo button is activated when color of the court is changed. Redo button is still disabled.
image
Click undo button until it goes back to the initial color of court, undo button is disabled.

### DIFF
--- a/src/components/NavBar/index.tsx
+++ b/src/components/NavBar/index.tsx
@@ -12,6 +12,8 @@ import LoginModalContent from "../Login";
 import { useEffect, useState } from "react";
 import { useStoreSelector } from "@/store/hooks";
 import { userInfo } from "os";
+import { ActionCreators } from "redux-undo";
+import { useDispatch } from "react-redux";
 
 const NavigationBar = () => {
   const { isOpen, onOpen, onClose } = useDisclosure();
@@ -44,6 +46,13 @@ const NavigationBar = () => {
     localStorage.removeItem("UserInfo");
     setLoginData(null);
   };
+  const dispatch = useDispatch();
+  const handleUndo = () => {
+    dispatch(ActionCreators.undo());
+  };
+  const handleRedo = () => {
+    dispatch(ActionCreators.redo());
+  };
 
   return (
     <Grid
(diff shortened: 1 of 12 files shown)


Jzeva (approved):

src/store/reducer/priceBarSlice.ts
```diff
@@ -0,0 +1,51 @@
...
+      state.blocks = action.payload;
+    },
+  },
+});
+
+export const { changeTileQuantity } = priceBarSlice.actions;
+
+export const priceBarData = (state: RootState) => state.priceBar;
+
+export default priceBarSlice.reducer;
```
tatoMa, line 51: I got confused. Is there any relationship between this file and undo and redo function?

src/components/NavBar/index.tsx
```diff
@@ -70,12 +81,15 @@ const NavigationBar = () => {
...
+            color={isThingsToUndo ? "white" : "#72818B"}
+            onClick={handleUndo}
           />
           <IconButton
             aria-label="Forward edit"
             icon={<RiArrowGoForwardLine />}
             variant="navbarIconBtn"
-            color="#72818B"
+            color={isThingsToRedo ? "white" : "#72818B"}
+            onClick={handleRedo}
```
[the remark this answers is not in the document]
jennawan: Thank you for the advice. Now using disabled attribute instead of color attribute.

src/store/reducer/priceBarSlice.ts
jennawan: No, there is no relationship between priceBarSlice and undo/redo function. I separated the priceBarSlice from tileSlice, and only add undo/redo function to tileSlice, as I only need the tileSlice to be undoable. If I do not separate them, when undoing the color of court, the redux-undo "future" will be overwritten by the update of pricebar state.

tatoMa: Make sense. Nice work.

tatoMa (approved):
lgtm

PR #115, opened 2022-07-11
Feature/cc 0094 replace tile calculator with tile data

jennawan:
For our main six types of courts, use mock data for the results showing in the price bar instead of tile calculator, as the calculator may cause a bug in price estimation. UseTileCalculation and tielNumberCalculator were remained in the repo, because they can be used for courts with user-customized sizes.

### DIFF
--- a/src/components/EditorSideBar/Blueprints.tsx
+++ b/src/components/EditorSideBar/Blueprints.tsx
@@ -46,7 +46,7 @@ const Blueprints: React.FC = () => {
     const courtSpec = mappedCourtSpecs.find((item: CourtSizeState) => item.courtId === courtId);
     dispatch(changeCourtSize(courtSpec));
 
-    const tileQtyOfSelectedCourt = mockTileData.find((item) => item.id === courtId)
+    const tileQtyOfSelectedCourt = mockTileData.find((item) => item.name === selectedCourt.name)
       ?.tileQty as AreaTileQty[];
     dispatch(changeCourtType(tileQtyOfSelectedCourt));
   };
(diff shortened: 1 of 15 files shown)


tatoMa (commented):
After the color of areas changes, if switch to a different court, the color remains. I think it is because of the mock color with the new approach "useTileCourt"? Anyway, just asking if it's a reasonable behavior or a bug.

jennawan:
> After the color of areas changes, if switch to a different court, the color remains. I think it is because of the mock color with the new approach "useTileCourt"? Anyway, just asking if it's a reasonable behavior or a bug.

@tatoMa Thank you for pointing that out. This "bug" is not due to the new approach useTileCount, but because of the colour was from tileSlice, and each area component in all courts such as "threePoint" in all courts, used the colour that in tileSlice. The group noticed it before, and we decided that it could be reasonable, as a user designed one court, he or she might have the same taste of colour style for every court.

tatoMa:
> > After the color of areas changes, if switch to a different court, the color remains. I think it is because of the mock color with the new approach "useTileCourt"? Anyway, just asking if it's a reasonable behavior or a bug.
> 
> @tatoMa Thank you for pointing that out. This "bug" is not due to the new approach useTileCount, but because of the colour was from tileSlice, and each area component in all courts such as "threePoint" in all courts, used the colour that in tileSlice. The group noticed it before, and we decided that it could be reasonable, as a user designed one court, he or she might have the same taste of colour style for every court.

Fair enough. Thanks for the explanation.

tatoMa (approved):
LGTM

Beryl-Zhou (approved):

PR #121, opened 2022-07-13
fix: cc-0088 price value NaN at uat and production

Beryl-Zhou:
fix: price display 0.00 at default page

### DIFF
--- a/src/components/PriceBar/TileColorBoard.tsx
+++ b/src/components/PriceBar/TileColorBoard.tsx
@@ -45,11 +45,10 @@ const TileColorBoard: React.FC = () => {
   };
 
   useEffect(() => {
-    if (data == undefined) return;
+    if (data === undefined) return;
     calculateTile();
     calculateDelivery();
     calculateInstallation();
-    console.log(priceDetails);
     const price = priceDetails.tilePrice + priceDetails.deliveryPrice + priceDetails.installPrice;
     const totalPrice = priceFormat(price);
     setTotalPrice(totalPrice);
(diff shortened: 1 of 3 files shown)


src/components/PriceBar/TileColorBoard.tsx
```diff
@@ -9,37 +9,51 @@ const TileColorBoard: React.FC = () => {
...
     if (courtList) {
-      installPrice += courtList.installationPrice / 100;
+      priceDetails.installPrice += courtList.installationPrice / 100;
     }
-    const price = tilePrice + deliveryPrice + installPrice;
-    // check price format, type transfer to string
+  };
+
+  useEffect(() => {
+    if (data == undefined) return;
```
Fuchih, line 48: Why you use "=="

src/components/PriceBar/TileColorBoard.tsx
```diff
@@ -9,37 +9,51 @@ const TileColorBoard: React.FC = () => {
...
+    calculateTile();
+    calculateDelivery();
+    calculateInstallation();
+    console.log(priceDetails);
```
Fuchih, line 52: Do we need to output a message to the console?

src/components/PriceBar/TileColorBoard.tsx
```diff
@@ -9,37 +9,50 @@ const TileColorBoard: React.FC = () => {
   const tileBlocks = useStoreSelector((state) => state.priceBar.blocks);
   const courts = useStoreSelector((state) => state.courtSize);
   const { data } = useGetPriceQuery(0);
-  const [useTotalPrice, setTotalPrice] = useState<string>("0.00");
+  const priceList = data?.find((item: IPriceCalculation) => !item.isDeleted);
+  const [useTotalPrice, setTotalPrice] = useState<string>("Loading...");
```
tatoMa, line 13: It is not acceptable to set a "loading" string into an apparent number variable. Please divide them into 2 variables and use conditions to render the loading text or price number.

Beryl-Zhou: actually, the final calculated result will be converted to string type to adjust the price format "X.X0"

PR #127, opened 2022-07-20
fix: small and medium court cannot change color

GitHubEason:


stevenliu986 (approved):
LGTM

Fuchih (approved):

tatoMa (approved):
LGTM
This is why using custom hooks with props pass-in is a better approach compare with copy and paste everywhere.

EmilyGong111 (approved):

PR #153, opened 2022-08-09
fix: the abnormal sign out button

GitHubEason:


stevenliu986 (approved):

Beryl-Zhou (approved):

tatoMa (commented):
As we discussed, the weird state behavior may be related to the useMemo hook. Try to find the solution instead of creating a new state. If cant, this implementation can be the final backup.

PR #171, opened 2022-09-13
Cc 0011 validate password when register

jennawan:
image
Add validation and error message for the password when a new user registers with the login modal.


Cecilia0109 (approved):
LGTM

src/components/Login/Register.tsx
```diff
@@ -122,6 +130,11 @@ const Register: React.FC<Props> = (props: Props) => {
               value={confirmPassword}
               onChange={(event) => setConfirmPassword(event?.currentTarget.value)}
             />
+            {weekPasswordMsg.length > 0 && (
```
kritomomo, line 133: Empty string is falsy, you can just use weekPasswordMsg &&

tatoMa (approved):
LGTM

PR #181, opened 2022-09-23
Cc 10 validate input value when changing the design name

jennawan:
image
Add validation for illegal design name (contains symbol or longer than 15 characters).
Refactor code of name check to custom hook and testing.

### DIFF
--- a/src/components/NavBar/EditorDesignName.tsx
+++ b/src/components/NavBar/EditorDesignName.tsx
@@ -15,9 +15,10 @@ import { useNameCheckFeedback } from "@/hooks/useNameCheckFeedback";
 
 const DesignName = () => {
   const designName = useStoreSelector((state) => state.courtSpecData.activeCourt.designName);
+  const nameList = useStoreSelector((state) => state.designName.nameList);
   const [newDesignName, setNewDesignName] = useState(designName);
   const { feedbackModalOpen, setFeedbackModalOpen, feedback, saveNameChange } =
-    useNameCheckFeedback(newDesignName);
+    useNameCheckFeedback(newDesignName, nameList);
 
   useEffect(() => {
     setNewDesignName(designName);
--- a/src/components/NavBar/EditorDesignName.tsx
+++ b/src/components/NavBar/EditorDesignName.tsx
@@ -10,7 +10,7 @@ import {
 import { BiPencil } from "react-icons/bi";
 import { useStoreSelector } from "@/store/hooks";
 import { useEffect, useState } from "react";
-import NameChangeFeedback from "./NameChangeFeedback";
+import NameChangeAlertModal from "./NameChangeAlertModal";
 import { useNameCheckFeedback } from "@/hooks/useNameCheckFeedback";
 
 const DesignName = () => {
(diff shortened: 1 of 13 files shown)


src/utils/checkName.ts
```diff
@@ -1,13 +1,17 @@
 const checkName = (name: string, nameList: string[]) => {
   let nameCheck = "passCheck";
   const checkStartEndSpace = /(^\s*)|(\s*$)/g;
+  // requirement for design name is only contain letter and number and less than 15 char
+  const validNamRegularExp = /^[A-Za-z0-9\s]{1,15}$/;
```
tatoMa, line 5: Improvement hits:
The max length and even the rules can be passed by props into this helper util so that makes this util DRY without making other similar helpers.

src/components/NavBar/EditorDesignName.tsx
```diff
@@ -8,49 +8,22 @@ import {
   EditableInput,
 } from "@chakra-ui/react";
 import { BiPencil } from "react-icons/bi";
-import { useDispatch } from "react-redux";
 import { useStoreSelector } from "@/store/hooks";
-import { changeDesignName } from "@/store/reducer/courtSpecDataSlice";
 import { useEffect, useState } from "react";
-import checkName from "@/utils/checkName";
 import NameChangeFeedback from "./NameChangeFeedback";
+import { useNameCheckFeedback } from "@/hooks/useNameCheckFeedback";
```
tatoMa, line 14: Do you think the name of these 2 component/custom hook are a little too similar? The first one is a modal maybe name it something modal?

src/utils/checkName.ts
jennawan: Thanks for the suggestion. The validation rules have been extracted to a validation function that passes to checkName. The validation rules can be configured using setValidation function.

src/components/NavBar/EditorDesignName.tsx
jennawan: Renamed NameChangeFeedback to NameChangeAlertModal.

tatoMa (approved):
Well Done!

src/utils/checkName.ts
```diff
@@ -1,13 +1,47 @@
...
-  }
-  if (name === "" || name.replace(checkStartEndSpace, "") === "") {
-    nameCheck = "blank";
-  }
-  // if () { nameCheck = "incorrect"; }
+  const onlyWordCharRegularExp = /^[a-zA-Z0-9\s]+$/g;
+  const validation = (name: string, nameList: string[]) => {
+    let result = "passCheck";
+    // check if name exists in nameList
+    if (nameList.includes(name) || nameList.includes(name.replace(checkStartEndSpace, ""))) {
```
ShawnWang86b, line 19: hi Jenna, a bit confused about this one, after checking AC we don't have to check the existing name,
but I found you write this function. and the  **const nameCheck = checkName(newDesignName, nameList, validation);
the nameList []** is blank, just double-check whether made this one or not on this ticket. thanks

jennawan: The feature of name existence checking was realised in the previous card. In this card, I refactored that part of code.
As for the nameList is blank, it is because the name list will only be updated when the user saves the court. It stands for the name list of saved courts by the current user.

ShawnWang86b (approved):

PR #182, opened 2022-09-23
Feature/cc 0005 show quotation detail

xunlutas:
add dropDownButton to show quotation detail


tatoMa (commented):
Any reason why still use mock data instead of our backend API?

src/components/ShoppingCart/dropDownButton.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  const scroll = "scroll";
+  const hidden = "hidden";
+  const noWrap = "nowrap";
+  const normal = "normal";
+  const clip = "clip";
+  const ellipsis = "ellipsis";
+
+  return (
+    <div
```
ZhihanChenZack, line 22: Since we use the Charka UI, should we limit the use of HTML tags like <div>?

src/components/ShoppingCart/dropDownButton.tsx
```diff
@@ -0,0 +1,59 @@
...
+      style={{
+        width: "100%",
+        display: "flex",
+        flexDirection: "row",
+        justifyContent: "center",
+        alignContent: "center",
+        overflow: "auto",
+      }}
+    >
```
ZhihanChenZack, line 31: From the initial shopping cart design, all text aline to the left, even if the content is short. e,g,
CartQuotationQuestion

But after the design of showing quotation detail, the content is centered. When the content is short the content will be shifted to the center of the container.

Cartdesignproblem2

It might be better to fix this problem.

ZhihanChenZack (commented):
Well done guys, I left some comments, and hope they can help you guys to improve the design.

src/components/ShoppingCart/dropDownButton.tsx
Evezzm: The text alignment was changed to justifyContent: "start". This branch was closed as it is renamed to cc-0008.

Evezzm: > From the initial shopping cart design, all text aline to the left, even if the content is short. e,g, CartQuotationQuestion
> 
> But after the design of showing quotation detail, the content is centered. When the content is short the content will be shifted to the center of the container.
> 
> Cartdesignproblem2
> 
> It might be better to fix this problem.

The text alignment was changed to justifyContent: "start". This branch was closed as it is renamed to cc-0008.

src/components/ShoppingCart/dropDownButton.tsx
Evezzm: Div tags were replaced by Box from Chakra UI.

PR #184, opened 2022-09-26
Feature/cc 0008 show quotation detail

xunlutas:
quotation_detail_UI

Drop Button for quotation detail


src/components/ShoppingCart/dropDownButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+type detail = {
+  detail: PriceBar[];
+};
+
+const DropDownButton = ({ detail }: detail) => {
+  const [show, setShow] = useState(false);
+
+  const handleToggle = () => setShow(!show);
+
+  const scroll = "scroll";
```
STY1997, line 16: In my opinion, you do not need to define these variables. Instead, you can simply use these strings as the chakra props.

src/components/ShoppingCart/dropDownButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const hidden = "hidden";
+  const noWrap = "nowrap";
+  const normal = "normal";
+  const clip = "clip";
+  const ellipsis = "ellipsis";
+
+  return (
+    <div
```
STY1997, line 24: I think you can use some chakra components (such as Box) instead of the div element.

src/styles/components/buttonStyles.ts
```diff
@@ -18,6 +18,17 @@ const buttonStyles = {
       _hover: { bg: "brand.secondary", opacity: "0.60" },
       _active: { bg: "brand.secondary", opacity: "0.60" },
     },
+    checkoutBtn: {
```
STY1997, line 21: Pls make sure that you pull the origin/main and merge your branch from the local/main first, before you start working.

STY1997 (commented):
Well done guys.

src/components/ShoppingCart/dropDownButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+          userSelect: "none",
+          whiteSpace: show ? normal : noWrap,
+          textOverflow: show ? clip : ellipsis,
+          textAlign: "center",
+        }}
+        data-testid="testShow"
+      >
+        {detail.map(
+          (content: PriceBar) => `Color:${content.color},  Quantity:${content.quantity},  `
+        )}
```
ZhihanChenZack, line 49: The detailed information can be put into a Box container and added with some necessary props instead of writing inline style in the Collapse component.

ZhihanChenZack (changes requested):
Well done guys!!

src/components/ShoppingCart/dropDownButton.tsx
```diff
@@ -0,0 +1,59 @@
...
+        style={{
+          overflowY: show ? scroll : hidden,
+          userSelect: "none",
+          whiteSpace: show ? normal : noWrap,
+          textOverflow: show ? clip : ellipsis,
+        }}
+        data-testid="testShow"
+      >
+        {content}
+      </Collapse>
```
Double-Twelve, line 50: Screen Shot 2022-09-26 at 7 40 04 pm


If the content of the quotation details can be fully displayed on the container, is there no need for the drop-down arrow?
<br/>
You can see my screenshot, when the quotation content can be displayed normally, I click on the drop-down arrow, the container will have a scroll bar, but it has no effect

src/styles/components/buttonStyles.ts
```diff
@@ -18,6 +18,17 @@ const buttonStyles = {
...
+      bg: "brand.secondary",
+      color: "white",
+      size: "lg",
+      fontSize: "lg",
+      fontWeights: "bold",
+      lineHeights: "22",
+      borderRadius: "6px",
+      _hover: { bg: "brand.secondary", opacity: "0.60" },
+      _active: { bg: "brand.secondary", opacity: "0.60" },
+    },
```
tatoMa, line 31: Seems the same as the shareBtn style on the same file. Duplicated code.

src/components/ShoppingCart/dropDownButton.tsx
xunlutas: Thank you for your advice. I have fixed that.

src/styles/components/buttonStyles.ts
xunlutas: OK, I have pulled the latest origin/main. Thank you.

src/components/ShoppingCart/dropDownButton.tsx
xunlutas: Thank you. I have fixed that.

src/components/ShoppingCart/dropDownButton.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { useState } from "react";
```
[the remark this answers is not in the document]
xunlutas: You are right. I have fixed that. Thank you.

src/components/ShoppingCart/dropDownButton.tsx
```diff
@@ -0,0 +1,65 @@
...
+  return (
+    <div
+      style={{
+        width: "100%",
+        display: "flex",
+        flexDirection: "row",
+        justifyContent: "center",
+        alignContent: "center",
+        overflow: "auto",
+      }}
```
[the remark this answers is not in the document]
xunlutas: Thank you. Fixed

src/components/ShoppingCart/dropDownButton.tsx
xunlutas: I have tried to put the content in Box. But some function will lose if I do that. In order to avoid too much inline style, I put all style property into const. 


 const collapseStyle: MotionStyle = {
          width: "100%",
          overflowY: show ? "scroll" : "hidden",
          userSelect: "none",
          whiteSpace: show ? "normal" : "nowrap",
          textOverflow: show ? "clip" : "ellipsis",
          textAlign: "center"
        }

 <Collapse
        startingHeight={25}
        in={show}
        style={collapseStyle}
        data-testid="testShow"
      >

It should be better now.

src/components/ShoppingCart/dropDownButton.tsx
xunlutas: It is the old one.  Maybe you should pull my latest branch.

src/styles/components/buttonStyles.ts
xunlutas: I think it is different. The background colour and size is very strange If I use shareBtn style for drop-down button.

ZhihanChenZack (approved):

tatoMa (approved):
LGTM

PR #185, opened 2022-09-26
Feature/cc 0013 remove item in shopping cart

PengchengYuGithub:
Add a delete confirm dialog when clicking delete button in shopping cart (before and after deleting):
image
image
"Cancel" will close the dialog and "delete" will remove correct cart item (the console log is just for showing effect):
image


src/components/ShoppingCart/CartListItem.tsx
```diff
@@ -4,12 +4,17 @@ import { FaPen } from "react-icons/fa";
...
 import { ICartItem } from "@/interfaces/cartItem";
 
-const CartListItem = (props: ICartItem) => {
+interface CartListItemProps {
+  item: ICartItem;
+  onDelete: (id: string) => void;
+}
+
+const CartListItem = ({ item, onDelete }: CartListItemProps) => {
   // import cart information
```
tatoMa, line 13: No need for comment. Coding should be self-documented themselves.

src/components/ShoppingCart/CartListItem.tsx
```diff
@@ -4,12 +4,17 @@ import { FaPen } from "react-icons/fa";
...
-  const productName = props.design.designName;
-  const quotation = props.quotation;
-  const quotationDetails = props.quotationDetails;
-  const image = props.previewPic;
+  const productName = item.design.designName;
+  const quotation = item.quotation;
+  const quotationDetails = item.quotationDetails;
+  const image = item.previewPic;
```
tatoMa, line 17: Do you know "destructuring assignment"? It can define several variables at once.

src/components/ShoppingCart/ShoppingCartContainer.tsx
```diff
@@ -53,11 +76,18 @@ const ShoppingCartContainer = ({ shoppingCart }: userCartList) => {
...
+                  setCartItemIdToDelete(id);
+                  onOpen();
+                }}
+                // id={cartRow.id}
+                // key={cartRow.user_id}
+                // user_id={cartRow.user_id}
+                // design={cartRow.design}
+                // quotation={cartRow.quotation}
+                // quotationDetails={cartRow.quotationDetails}
+                // previewPic={cartRow.previewPic}
```
tatoMa, line 90: Don't leave anything like that. If you need to do it in the future, add "Todo" as the start of the comment.

src/components/ShoppingCart/CartListItem.tsx
PengchengYuGithub: Got it. These comments will not be reserved in the code.

src/components/ShoppingCart/CartListItem.tsx
PengchengYuGithub: We simply follows others style and actually we should use destructuring assignment.

src/components/ShoppingCart/ShoppingCartContainer.tsx
PengchengYuGithub: These are original codes from cc06 and we comment them in case they want to restore; we now understand it is not a good behavior.

tatoMa (approved):
LGTM

GLC-coder (approved):
LGTM, Good job!

PR #186, opened 2022-09-27
fix: change Chakra style acronym to full name on cc-08 show quotation…

xunlutas:
change Chakra style acronym to full name on cc-08 show quotation


tatoMa (approved):
LGTM

GLC-coder (approved):
LGTM

PR #187, opened 2022-09-27
test: cc-0022 delete the invalidate designApi.test.ts and incluede de…

GLC-coder:
test/cc-0022 refactor unit testing in designApi has been finished and it is ready to review. As we don't need to conduct unit testing for designApi created via RTK Query, it has been included in the collectCoverageFrom in jest.config.js.
image


RedRe4per (approved):
LGTM

tatoMa (approved):
LGTM

PR #190, opened 2022-09-29
Feature cc39 resolve login conflict

zhuocun:
image

Context: a user have registered an account with email, and then tries to login with google with the same email.
Feature: a new window will show and indicate user to connect account with google, after that, user can login with both way.

### DIFF
--- a/src/components/Login/SelectLogin.tsx
+++ b/src/components/Login/SelectLogin.tsx
@@ -33,10 +33,11 @@ export interface GoogleLoginRes {
   firstName: string;
   lastName: string;
   isActivated: string;
+  otp: string;
   needConnection: boolean;
 }
 
-export default function SelectLogin(props: Props) {
+const SelectLogin = (props: Props) => {
   const dispatch = useDispatch();
   const { onClose, updateLoginData, initialRef, nextStep, connectionStep } = props;
 
@@ -120,4 +121,6 @@ export default function SelectLogin(props: Props) {
       </ModalFooter>
     </>
   );
-}
+};
+
+export default SelectLogin;
--- a/src/components/Login/SelectLogin.tsx
+++ b/src/components/Login/SelectLogin.tsx
@@ -21,7 +21,7 @@ import { updateUserInfo } from "@/store/reducer/userSlice";
 
 interface Props {
   onClose: () => void;
-  updateLoginData: (data: any) => void;
+  updateLoginData: (data: GoogleLoginRes) => void;
   initialRef: React.LegacyRef<HTMLButtonElement> | undefined;
   nextStep: () => void;
   connectionStep: (existedUserInfo: GoogleLoginRes) => void;
--- a/src/components/Login/SelectLogin.tsx
+++ b/src/components/Login/SelectLogin.tsx
@@ -15,7 +15,7 @@ import { FaEnvelope } from "react-icons/fa";
 import { FcGoogle } from "react-icons/fc";
 import { useGoogleLogin } from "@react-oauth/google";
 import React from "react";
-import { api } from "../../utils/axios";
+import { api } from "@/utils/axios";
 import { useDispatch } from "react-redux";
 import { updateUserInfo } from "@/store/reducer/userSlice";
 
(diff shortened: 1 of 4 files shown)


src/components/Login/accountConnection/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+        localStorage.setItem("UserInfo", JSON.stringify(data));
+        dispatch(updateUserInfo(data));
+        updateLoginData(data);
+        setStep(1);
+        onClose();
+      }
+    } catch (err) {
+      console.warn(err);
+    }
+  };
```
tatoMa, line 53: Are we going to have this connection API called only here or maybe somewhere else in the future? If only here then all good, otherwise placing it into redux might be a better way. Just asking.

tatoMa (approved):
All good.

src/components/Login/accountConnection/index.tsx
zhuocun: Thank you 桃总, for now I haven't come up with a situation where this call is also needed, I'll move it to Redux if there's a need for it in the future :).

zhuocun: > Are we going to have this connection API called only here or maybe somewhere else in the future? If only here then all good, otherwise placing it into redux might be a better way. Just asking.

Thank you 桃总, for now I haven't come up with a situation where this call is also needed, I'll move it to Redux if there's a need for it in the future :).

src/components/Login/SelectLogin.tsx
```diff
@@ -37,13 +51,16 @@ export default function SelectLogin(props: Props) {
       method: "post",
       requestData: codeResponse,
     });
+    const googleLoginRes: GoogleLoginRes = data;
     try {
-      if (data) {
+      if (!googleLoginRes.needConnection) {
         // Store user data into local storage after logging
         localStorage.setItem("UserInfo", JSON.stringify(data));
```
Chanran, line 58: use "googleLoginRes" variable？

src/components/Login/accountConnection/index.tsx
Chanran: You can leave a comment to describe it just in case.

src/components/Login/accountConnection/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+const AccountConnection: React.FC<Props> = ({
+  existedUserInfo,
+  setStep,
+  updateLoginData,
+  onClose,
+}) => {
+  const dispatch = useStoreDispatch();
+  const onConnect = async () => {
+    const { data } = await api("/user/connect", {
```
Chanran, line 35: Is there possible to add interface to describe "data" variable？

src/components/Login/accountConnection/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+        googleId: existedUserInfo?.googleId,
+        email: existedUserInfo?.email,
+      },
+    });
+    try {
+      if (data) {
+        localStorage.setItem("UserInfo", JSON.stringify(data));
+        dispatch(updateUserInfo(data));
+        updateLoginData(data);
+        setStep(1);
```
Chanran, line 47: Maybe you should define step number as enum cause reviewers or other developers don't know what "step 1" means. What do you think？ @tatoMa

Chanran: But this could be solved in next pr.No blocks in this pr.

src/components/Login/SelectLogin.tsx
zhuocun: > use "googleLoginRes" variable？

Since the type of response is AxiosResponse<any, any>, I tried to use feature of TS to handle response.data, so I assigned it to googleLoginRes.

Chanran: Yes I know, I think you should replace the "data" by the "googleLoginRes" variable to do the localStorage setItem at line 58 ~ line 60.

zhuocun: > Yes I know, I think you should replace the "data" by the "googleLoginRes" variable to do the localStorage setItem at line 58 ~ line 60.

Got it! :)

src/components/Login/accountConnection/index.tsx
zhuocun: > But this could be solved in next PR.No blocks in this PR.

understood! Thanks for reminding me of that!

src/components/Login/accountConnection/index.tsx
zhuocun: > Is there possible to add interface to describe "data" variable？

Yes, it is. I could handle it in the same way as in "SelectLogin.ts" to assign it to a variable like "googleLoginRes" to give it strong type :)

jennawan (approved):
LGTM!

PR #201, opened 2022-10-03
fix: the password entered during registration contains at least one s…

Double-Twelve:
### The password should contain at least one special character

Before:
截屏2022-09-29 下午11 02 04

After:
Snipaste_2022-10-03_18-17-27


tatoMa (approved):
LGTM

GLC-coder (approved):
Well done!LGTM!

PR #202, opened 2022-10-04
Feature/cc38 - button slice refactoring

wwwseven:
Delete some slices of button and put them in one slice;

Refactor some button callback functions;


Double-Twelve:
**When the user clicks the shopping cart button without logging in, a login prompt box will pop up, but the sidebar is not hidden**

Snipaste_2022-10-04_16-58-44

Double-Twelve:
### I'm not sure if the following two points should be completed

- These buttons, which are framed in red in my screenshot, should also hide the sidebar when they are clicked

Snipaste_2022-10-04_16-58-44

- When I click on the paint bucket, instead of selecting the color, I click on the sidebar, the text above the paint bucket will always be there, I'm not sure if this should be hided

Snipaste_2022-10-04_16-58-44

Snipaste_2022-10-04_17-12-29

Double-Twelve (approved):
Well Done!

src/store/index.ts
```diff
@@ -2,20 +2,18 @@ import { configureStore, ThunkAction, Action } from "@reduxjs/toolkit";
 import userReducer from "./reducer/userSlice";
 import tileReducer from "./reducer/tileSlice";
 import courtColorReducer from "./reducer/courtColorSlice";
-import rulerControlReducer from "./reducer/rulerControlSlice";
+import designPageButtonReducer from "./reducer/designPageButtonSlice";
```
tatoMa, line 5: The name of this file can be more generic or specific. The logic in this slice is all boolean which means toggles, and not all of these booleans are on design page( there is one shopping cart toggle)

tatoMa:
It is so nice that you put the sidebar toggle into the new slice file. Nice work overall.

src/store/reducer/buttonToggleSlice.ts
```diff
@@ -0,0 +1,78 @@
...
+      };
+    },
+    usePaintBucket: (state: ButtonToggleState, action: PayloadAction<boolean>) => {
+      return {
+        ...state,
+        isPaintPopoverOpen: action.payload,
+        isSideBarOpen: false,
+      };
+    },
+    useSavePopover: (state: ButtonToggleState, action: PayloadAction<boolean>) => {
```
tatoMa, line 45: The prefix "use" normally only uses in the name of hooks and these are redux dispatches. You can simply change the name of them to something like "switchSomething" or "toggleSomething" just an example. Ultimately, we want to keep all of them consistent.

src/store/reducer/buttonToggleSlice.ts
```diff
@@ -0,0 +1,78 @@
...
+      };
+    },
+    switchSideBar: (state: ButtonToggleState, action: PayloadAction<boolean>) => {
+      return {
+        ...state,
+        isSideBarOpen: action.payload,
+        isCartOpen: false,
+      };
+    },
+    useLoginModal: (state: ButtonToggleState, action: PayloadAction<boolean>) => {
```
tatoMa, line 59: Same as above.

tatoMa (approved):
LGTM

PR #204, opened 2022-10-05
Cc 44 sign up user flow improvement

xunlutas:
image
In verify step, users are not allowed to go back and the data will be removed when exit


src/components/Login/ModalOperater/index.tsx
```diff
@@ -1,21 +1,32 @@
...
-        color="black"
-        fontSize="28px"
-        onClick={prevStep}
-      ></IconButton>
+    <Flex
+      width="100%"
+      flexDirection="row"
+      justifyContent={currentStep !== 4 ? "space-between" : "end"}
+    >
+      {currentStep !== 4 ? (
```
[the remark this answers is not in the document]
xunlutas: In my opinion, writing a single individual ModalOperator only for email verification is not reusable. Maybe there is a better solution for the issue.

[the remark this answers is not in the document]
zhuocun: > In my opinion, writing a single individual ModalOperator only for email verification is not reusable. Maybe there is a better solution for the issue.

That makes sense :) I reckon it is based on whether in the future we will still need the modal operator to accept "current step" parameters, if not, then in fact we need to insert "current step" for no use each time we use the modal operator

zhuocun (commented):
Thanks for inviting me! Everything looks good, but I guess I need 桃总’s opinion about the solution issue we’ve talked about before, and there will be no problem for me if 桃总 thinks the solution is ok!

src/components/Login/index.tsx
```diff
@@ -55,6 +55,7 @@ const LoginModalContent = (props: Props) => {
       case 2:
```
tatoMa, line 55: Do you think using an object or enum instead of step 1-6 is a better idea and easier to debug later?

tatoMa: As far as I read. There are several components using these steps by passed props. Using these numbers makes debugging harder than an object or enum.

xunlutas: image
I have used enum for stepName. I think it is clearer than just using a number.

src/components/Login/ModalOperater/index.tsx
```diff
@@ -4,17 +4,17 @@ import useAuthRequest from "../helpers/authRequest";
...
+  currentStep: string;
 };
 
 const ModalOperator: React.FC<Props> = ({ handleCloseModal, prevStep, currentStep }) => {
   return (
     <Flex
       width="100%"
       flexDirection="row"
-      justifyContent={currentStep !== 4 ? "space-between" : "end"}
+      justifyContent={currentStep !== "EmailVerification" ? "space-between" : "end"}
```
tatoMa, line 15: The String "EmailVerification" can be imported from the enum "stepName" you have just created from Login/index file. Normally, we do not use any plain string in any file like this.

tatoMa (approved):

PR #207, opened 2022-10-05
Fix/cc 66 account conflicts

zhuocun:
0. Please review only the first 3 commits in the "Commits" section, since the "Files changed" section contains all changes from CC-38 that I merged and resolved conflicts from the main branch. The component where I resolved conflicts is "NavBar".

I'm really not sure if I should merge the main with updates to my branch before the code review is finished since it seems like it increases the number of file changes which makes code review harder :(. Please let me know if you think I shouldn't merge the main and I'll drop the commits.

1. This issue is to resolve the situation that: 
In the previous version, the front end regards accounts signed up with email and google by the same email as two separate accounts. If users log in with google, the front end will send the request by google id, or else by user id.

However, in our latest user account logic, the front end should regard accounts in the situation above as one same account, since they are connected.

2. this issue update UI of the modal of the account connection component based on requests from BA.
image


Double-Twelve (approved):
Well done!

PR #208, opened 2022-10-07
Fix/cc 66 resolve account conflicts

zhuocun:
1. This issue is to resolve the situation that: 
In the previous version, the front end regards accounts signed up with email and google by the same email as two separate accounts. If users log in with google, the front end will send the request by google id, or else by user id.

However, in our latest user account logic, the front end should regard accounts in the situation above as one same account, since they are connected.

2. this issue update UI of the modal of the account connection component based on requests from BA.
image


Double-Twelve (approved):
Well done!

tatoMa (approved):
LGTM

PR #210, opened 2022-10-10
Bugfix/cc 0070 replace chakra UI/icons with react icons

Evezzm:
1. Removed historical chakra-ui/icons by replacing them with react icons
2. Update react icons styles to match the previous styles

### DIFF
--- a/src/components/TopBar/index.tsx
+++ b/src/components/TopBar/index.tsx
@@ -10,7 +10,7 @@ import {
   Grid,
   Tooltip,
 } from "@chakra-ui/react";
-import { TriangleUpIcon } from "@chakra-ui/icons";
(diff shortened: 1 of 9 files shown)


src/components/TopBar/index.tsx
```diff
@@ -217,7 +218,8 @@ const TopBar = () => {
               color="brand.primary"
               border="none"
               marginTop={3}
-              as={TriangleUpIcon}
+              as={RiArrowUpSFill}
+              size={25}
```
tatoMa, line 222: Might need a negative margin or anything to place this icon at the right place.
image

tatoMa (approved):
LGTM

PR #212, opened 2022-10-11
fix: reset failure before download

Double-Twelve:
Before fixed, court design has not been reset and relocated at the center of the downloaded PDF. The reason is the canvas rendering layer is **asynchronous**. Even the code has set to reset the size of the court first, and then execute the download function, the image in the downloaded PDF is still the image that has not been reset.  

I didn't figure out a better way to avoid this, I just set a timer to perform the download function 0.5 second after resetting the size.


tatoMa (approved):
It is not the best solution but I haven't found a better way to solve it yet. So good for now.

zhuocun (approved):
An intelligent and time-saving method :)

PR #218, opened 2022-10-18
Feature/cc 35 share users design as a template

ZhihanChenZack:
**Adding sharing template function.**

- Template court name (different from the court name) with a length verification.
- Template description with a length verification.
- Publisher name display.
- Long username handle (e.g. 'supper username' to 'supper us...').
- Create templateApi.

image

**Error handling:**
- Court name empty 
image
- Court name over length
image
- Description over length
image

**Successful publish**
image


tatoMa (commented):
image

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+  courtNameFullErr: false,
+  courtNameNullErr: false,
+  descriptionOverLimit: false,
+};
+
+const inputErrorMsg = {
+  nameFullErrMsg: `Court name cannot have more than ${maxCourtNameLen} characters`,
+  nameNullErrMsg: "Court name cannot be empty",
+  descriptionLenErrMsg: `Description can not over ${maxDescriptionLen} words`,
+};
```
kritomomo, line 49: It would be better to move these to a constant file

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+            <FormControl>
+              <FormLabel>Court Preview:</FormLabel>
+              <Box width={300} height={150} backgroundColor="orange" margin="30px auto">
+                Court Image
+              </Box>
+            </FormControl>
+            <Flex gap="24px" justifyContent="center" flexWrap="wrap">
+              <Badge margin="16px" colorScheme="green">
+                {courtType}
+              </Badge>
```
kritomomo, line 149: const courtType = "basketball";
Why not just put this basketball here?

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+  const courtNameRef = useRef<HTMLInputElement>(null);
+  const descriptionRef = useRef<HTMLTextAreaElement>(null);
+  const [inputError, setInputError] = useState(inputErrorInit);
+  const [textAreaLen, setTextAreaLen] = useState(0);
+  const { userId, firstName, lastName } = useStoreSelector((state) => state.user);
+  const { activeCourt: selectedCourt } = useStoreSelector((state) => state.courtSpecData);
+  const { court: selectedCourtTileData } = useStoreSelector((state) => state.tile.present);
+  const [addTemplate] = useAddTemplateMutation();
+  const courtType = "basketball";
+  const userFullName = userNameEllip(`${firstName} ${lastName}`, maxUserNameDisplay);
```
kritomomo, line 61: Is the requirement here limited to display only 20 characters, or the part that exceeds the width will not be displayed, because this can be implemented with css

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+                onChange={handleTextAreaLenChange}
+                resize="none"
+                ref={descriptionRef}
+              />
+            </FormControl>
+            <Text role="wordCount" color={textAreaLen < maxDescriptionLen ? "black" : "crimson"}>
+              {inputError.descriptionOverLimit
+                ? inputErrorMsg.descriptionLenErrMsg
+                : `${textAreaLen}/${maxDescriptionLen} words`}
+            </Text>
```
kritomomo, line 206: What is the purpose of using FormControl here?
image

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+                width="20rem"
+                height="11rem"
+                backgroundColor="orange"
+                justifyContent="center"
+                alignItems="center"
+                margin="1.8rem auto"
+              >
+                Court Image
+              </Flex>
+            </FormControl>
```
kritomomo, line 145: Why you use `FormControl` here

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+    const textAreaLength = e.currentTarget.value
+      .replace(regex, " ")
+      .split(" ")
+      .filter((item) => item != "").length;
+    setTextAreaLen(textAreaLength);
+    setInputError((inputError) => ({
+      ...inputError,
+      descriptionOverLimit: textAreaLength > maxDescriptionLen,
+    }));
+  };
```
kritomomo, line 93: If the special symbols are not counted here, then if I enter 201 @, it will not count as exceeding the word count requirement?

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,228 @@
...
+              color="crimson"
+              visibility={
+                inputError.courtNameFullErr || inputError.courtNameNullErr ? "visible" : "hidden"
+              }
+              fontSize="0.8rem"
+            >
+              {inputError.courtNameFullErr
+                ? inputErrorMsg.nameFullErrMsg
+                : inputErrorMsg.nameNullErrMsg}
+            </Text>
```
kritomomo, line 191: Maybe you can wrap this part into an error message component

src/components/CreateTemplate/index.tsx
STY1997: Yes, we only count the normal punctuations which we can directly input from keyboard. If user type some special symbols other than our regex, they will be counted as normal characters.

STY1997: > If the special symbols are not counted here, then if I enter 201 @, it will not count as exceeding the word count requirement?

Yes, we only count the normal punctuations which we can directly input from keyboard. If user type some special symbols other than our regex, they will be counted as normal characters.

src/components/CreateTemplate/index.tsx
ZhihanChenZack: Considering the template tags for courtType currently only have the basketball type(tennis or mixed court is added in the future). We might need to read the court type from the design specifications in the future. This const for courtType is just a temporary solution for the court type tag.

src/components/CreateTemplate/index.tsx
ZhihanChenZack: Deleted

src/components/CreateTemplate/index.tsx
ZhihanChenZack: Deleted

src/redux/api/templateApi.ts
```diff
@@ -0,0 +1,66 @@
...
+      }),
+      invalidatesTags: (_result, _err, arg) => [{ type: "template", id: arg }],
+    }),
+
+    updateTemplate: builder.mutation<ITemplateDataDb, Partial<ITemplate> & Pick<ITemplate, "_id">>({
+      query: (templateid) => ({
+        url: `/templates/${templateid}`,
+        method: "PUT",
+      }),
+      invalidatesTags: (_result, _err, arg) => [{ type: "template", id: arg._id }],
```
GLC-coder, line 55: In my view, it is better to add {type:"template", id:"LIST"} after { type: "template", id: arg._id }. When a specified template was updated, then the cache of the template with that specified id was affected, but the "LIST" of all templates in the template community was not updated. Users need to reload the template community page by themselves if they want to see the updated template which they updated just a few seconds ago. Add {type:"template", id:"LIST"} here, as well as at the providesTags at getTemplates endpoint can solve this issue.

src/redux/api/templateApi.ts
```diff
@@ -0,0 +1,66 @@
...
+      }),
+      invalidatesTags: ["template"],
+    }),
+
+    deleteTemplate: builder.mutation<boolean, string>({
+      query: (templateId) => ({
+        url: `/templates/${templateId}`,
+        method: "DELETE",
+      }),
+      invalidatesTags: (_result, _err, arg) => [{ type: "template", id: arg }],
```
GLC-coder, line 47: Same here, it is my suggestion that you can add {type:"template", id:"LIST"} after { type: "template", id: arg._id }

src/redux/api/templateApi.ts
```diff
@@ -0,0 +1,66 @@
...
+        result ? [{ type: "template", id: result._id }] : ["template"],
+    }),
+
+    addTemplate: builder.mutation<ITemplateDataDb, Omit<ITemplate, "_id">>({
+      query: (newTemplate) => ({
+        url: "templates",
+        method: "POST",
+        body: newTemplate,
+      }),
+      invalidatesTags: ["template"],
```
GLC-coder, line 39: Same here, it is my suggestion that you can add {type:"template", id:"LIST"} after { type: "template", id: arg._id }

src/redux/api/templateApi.ts
```diff
@@ -0,0 +1,66 @@
...
+    getTemplates: builder.query<Omit<ITemplateDataDb, "__v" | "isDeleted">[], string>({
+      query: (userId) => `/templates?user_id=${userId}`,
+      transformResponse: (result: ITemplateDataDb[], _meta, _arg) => {
+        return result.map((item) => _.omit(item, ["__v", "isDeleted"]));
+      },
+      providesTags: (result, _err, _arg) =>
+        result
+          ? [...result.map(({ _id }) => ({ type: "template" as const, id: _id })), "template"]
+          : ["template"],
+    }),
```
GLC-coder, line 22: Same here, it is my suggestion that you can add {type:"template", id:"LIST"} after { type: "template", id: arg._id }

GLC-coder (commented):
Well done! TQL

src/redux/api/templateApi.ts
STY1997: > In my view, it is better to add {type:"template", id:"LIST"} after { type: "template", id: arg._id }. When a specified template was updated, then the cache of the template with that specified id was affected, but the "LIST" of all templates in the template community was not updated. Users need to reload the template community page by themselves if they want to see the updated template which they updated just a few seconds ago. Add {type:"template", id:"LIST"} here, as well as at the providesTags at getTemplates endpoint can solve this issue.

I defined  [{ type: "template", id: _id_1}, {type: "template", id: _id_2}, {type: "template", id: _id_3}, ... ... ] rather than {type: "template" id: "LIST"} in useQuery endpoints. It has the same effect as what you said. For example, when {type: "template", id: 123456} is invalid in the mutation endpoints, it will do the re-fetch action in useQuery which contain the tag {type: "template", id: 123456}.

GLC-coder: 👍

src/__tests__/components/CreateTemplate/SuccessNotice.test.tsx
```diff
@@ -0,0 +1,32 @@
+import { screen, waitFor } from "@testing-library/react";
+import SuccessNotice from "@/components/CreateTemplate/SuccessNotice";
+import renderWithMockedProvider from "../../utils";
+import userEvent from "@testing-library/user-event";
+import React from "react";
+
+describe("SuccessNotice", () => {
+  test("should render successNotice page when user punish template successfully", () => {
```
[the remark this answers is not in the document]
Cecilia0109: > "it" and "test" have the same effect, but I suggest it is better to use the more semantic word, "It should" is more reasonable than "test should". Another point is to be consistent, because I found that some test files are it and some are test, which is not really good

Thank you!  have replaced "test" in all test files

Double-Twelve:
Pro Half Court and Medium Court cannot be shared

STY1997:
> Pro Half Court and Medium Court cannot be shared

This is caused by missing the court category information of these two courts...I have fixed this issue and re-pull request the branch in the backend repository. It will be solved after the new fixed backend branch merge to the main.

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+              placeholder={`Description: maximum ${MAX_DESCRIPTION_LEN} words`}
+              onChange={handleTextAreaLenChange}
+              resize="none"
+              ref={descriptionRef}
+            />
+            <Text role="wordCount" color={textAreaLen < MAX_DESCRIPTION_LEN ? "black" : "crimson"}>
+              {inputError.descriptionOverLimit
+                ? INPUT_ERROR_MSG.descriptionLenErrMsg
+                : `${textAreaLen}/${MAX_DESCRIPTION_LEN} words`}
+            </Text>
```
Double-Twelve, line 188: If you delete _FormControl_, _FormLabel_ can also be removed.

However, I think Robert means that if you use _FormControl_, you can use _FormErrorMessage_, you don't need to write _Text_ anymore, you can refer to the example he provided

src/__tests__/components/CreateTemplate/ErrorMsg.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { ITemplateErrorInput } from "@/interfaces/template";
+
+describe("ErrorMsg", () => {
+  it("should not display error message when no validations occur", () => {
+    const userInputError: ITemplateErrorInput = {
+      courtNameFullErr: false,
+      courtNameNullErr: false,
+      descriptionOverLimit: false,
+    };
+    const inputErrorMsg = INPUT_ERROR_MSG;
```
tatoMa, line 13: You can use beforeEach to store all duplicated test cases in it. The last one of your test case only needs to modify a little and will make your test file cleaner if you want to.

src/__tests__/components/CreateTemplate/SuccessNotice.test.tsx
```diff
@@ -0,0 +1,28 @@
+import { screen, waitFor } from "@testing-library/react";
+import SuccessNotice from "@/components/CreateTemplate/SuccessNotice";
+import renderWithMockedProvider from "../../utils";
+import userEvent from "@testing-library/user-event";
+import React from "react";
+
+describe("SuccessNotice", () => {
+  it("should render successNotice page when user punish template successfully", () => {
+    renderWithMockedProvider(<SuccessNotice isOpen={true} onClose={() => void {}} />);
```
tatoMa, line 9: The only different between renderWithMockedProvider and render(the one from @testing-library/react) is we wrapped redux in renderWithMockedProvider but no redux in the render from the testing library.

So in this case, the render from the testing library should do all these works.

src/__tests__/components/CreateTemplate/index.test.tsx
```diff
@@ -0,0 +1,74 @@
+import CreateTemplate from "@/components/CreateTemplate";
+import renderWithMockedProvider from "../../utils";
+import { fireEvent, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+describe("Create Template", () => {
+  it("Should render all page titles", () => {
+    renderWithMockedProvider(<CreateTemplate isOpen={true} onClose={() => void {}} />);
```
tatoMa, line 8: This is the one definitely needs redux so renderWithMockedProvider is the best for this case.

src/components/CreateTemplate/SuccessNotice.tsx
```diff
@@ -0,0 +1,39 @@
...
+        ml="98px"
+        flexDirection="column"
+        alignItems="center"
+        justifyContent="center"
+      >
+        <ModalHeader mt="15px" mb="-10px" color="brand.primary" fontSize="md">
+          Congratulations You have successfully submitted a new template!
+        </ModalHeader>
+        <ModalFooter>
+          <Button bg="button.hover" color="fontcolor.primary" onClick={onClose} role="closeBtn">
```
tatoMa, line 31: Give me a reason why change the value of the role property.

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+            <Flex>
+              <FormControl
+                width="50%"
+                marginTop="1rem"
+                isRequired
+                isInvalid={inputError.courtNameFullErr}
+              >
+                <FormLabel marginBottom="1rem">Template Court Name:</FormLabel>
+                <Input
+                  role="courtNameInput"
```
tatoMa, line 149: Same, why change the role value.

src/components/CreateTemplate/SuccessNotice.tsx
```diff
@@ -0,0 +1,39 @@
...
+    <Modal isOpen={isOpen} onClose={onClose} size="2xl" isCentered>
+      <ModalOverlay />
+      <ModalContent
+        display="flex"
+        ml="98px"
+        flexDirection="column"
+        alignItems="center"
+        justifyContent="center"
+      >
+        <ModalHeader mt="15px" mb="-10px" color="brand.primary" fontSize="md">
```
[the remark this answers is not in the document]
STY1997: marginBottom is used to reduce the spacing between modalHead and modalFooter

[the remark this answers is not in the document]
STY1997: > 1. Please use full names of the attributes
> 2. why use mb? what is it purpose?

marginBottom is used to reduce the spacing between modalHead and modalFooter

src/components/CreateTemplate/SuccessNotice.tsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+const SuccessNotice = (prop: Props) => {
+  const { isOpen, onClose } = prop;
+  return (
+    <Modal isOpen={isOpen} onClose={onClose} size="2xl" isCentered>
+      <ModalOverlay />
+      <ModalContent
+        display="flex"
+        ml="98px"
```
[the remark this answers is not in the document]
STY1997: > Please use full names of the attributes _marginLeft="98px"_

fixed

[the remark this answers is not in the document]
STY1997: > Please use full names of the attributes _marginLeft="98px"_

fixed

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+    setTextAreaLen(0);
+    onClose();
+  };
+
+  const checkNameLength = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const nameInputLen = e.currentTarget.value.length;
+    setInputError((inputError) => ({
+      ...inputError,
+      courtNameFullErr: nameInputLen > MAX_COURTNAME_LEN,
+      courtNameNullErr: nameInputLen === 0,
```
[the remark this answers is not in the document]
STY1997: > Should remove whitespace from both ends before getting the length of name, or first determine whether the name are all whitespace

fixed, thanks for advising

src/components/CreateTemplate/index.tsx
STY1997: > If you delete _FormControl_, _FormLabel_ can also be removed.
> 
> However, I think Robert means that if you use _FormControl_, you can use _FormErrorMessage_, you don't need to write _Text_ anymore, you can refer to the example he provided

fixed

src/components/CreateTemplate/SuccessNotice.tsx
STY1997: > Give me a reason why change the value of the role property.

aria-labels are used rather than customised roles now

src/__tests__/components/CreateTemplate/index.test.tsx
STY1997: > This is the one definitely needs redux so renderWithMockedProvider is the best for this case.

fixed

src/__tests__/components/CreateTemplate/SuccessNotice.test.tsx
STY1997: > The only different between renderWithMockedProvider and render(the one from @testing-library/react) is we wrapped redux in renderWithMockedProvider but no redux in the render from the testing library.
> 
> So in this case, the render from the testing library should do all these works.

fixed

src/__tests__/components/CreateTemplate/ErrorMsg.test.tsx
STY1997: > You can use beforeEach to store all duplicated test cases in it. The last one of your test case only needs to modify a little and will make your test file cleaner if you want to.

beforeEach() is applied to initialised the test data before each test now

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+            <Button
+              role="publishBtn"
+              colorScheme="blue"
+              variant="shareBtn"
+              paddingX="3rem"
+              onClick={submitTemplate}
+            >
+              Publish
+            </Button>
+            <Button role="cancelBtn" fontSize="18px" onClick={closeWindow} paddingX="3rem">
```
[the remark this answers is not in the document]
Cecilia0109: > fontSize should be "lg" Please add borderRadius: "6px"

Fixed thanks

src/components/CreateTemplate/SuccessNotice.tsx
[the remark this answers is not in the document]
Cecilia0109: > The background color of the button after the hover is very similar to the text color, so it is not easy to see clearly
> 
> image

Fixed thanks

src/components/CreateTemplate/index.tsx
```diff
@@ -0,0 +1,210 @@
...
+                    whiteSpace="nowrap"
+                    overflow="hidden"
+                    textOverflow="ellipsis"
+                  >
+                    {`${firstName} ${lastName}`}
+                  </Text>
+                </Flex>
+              </Box>
+            </Flex>
+            <ErrorMsg userInputError={inputError} inputErrorMsg={INPUT_ERROR_MSG} />
```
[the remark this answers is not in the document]
Cecilia0109: > The statement should be wrapped FormControl and set width the same as the input, otherwise if the error message is too long, it will be displayed in the publisher's area
> 
> image

Fixed thanks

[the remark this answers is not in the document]
Cecilia0109: > The statement should be wrapped FormControl and set width the same as the input, otherwise if the error message is too long, it will be displayed in the publisher's area
> 
> image

Fixed thanks

GLC-coder (approved):
Great work, LGTM

src/__tests__/components/CreateTemplate/index.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+    const descriptionInput = screen.getByRole("textbox", { name: "textArea" });
+    fireEvent.change(descriptionInput, {
+      target: { value: "This my bag" },
+    });
+    await waitFor(() => {
+      expect(screen.getByText("11/200 letters")).toBeVisible();
+    });
+  });
+});
```
kritomomo, line 74: I see that both fireEvent and userEvent are used here, but both of them have click and input methods. I recommend using userEvent
[common-mistakes-with-react-testing-library](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library)
image

ZhihanChenZack: All fireEvent has been changed to userEvent.

PR #221, opened 2022-10-25
refactor: make error messages consistent when design name is incorrect

Double-Twelve:
### The middle of the top bar can directly modify the name of the design
image

### The design name can be changed e by clicking "Save As"
image

**Before updated:**
These two input boxes have methods for validating user input respectively, and the error messages that are prompted are also different.

**After updated:**
Use the same validation method.


wwwseven (approved):
LGTM

src/utils/checkDesignName.ts
```diff
@@ -0,0 +1,18 @@
+import { DESIGN_NAME_REGULAR_EXPRESSION } from "@/constants/courtData";
+
+const checkDesignName = (name: string, nameList: string[]) => {
+  // when canBeBlank is false, check if name is blank
+  if (name.length === 0) {
```
tatoMa, line 5: Might need the trim here. I see you have trim when you call checkDesignName else where every time.

src/components/TopBar/SaveBoard.tsx
```diff
@@ -43,12 +40,10 @@ const SaveBoard: React.FC = () => {
   const designsData = useStoreSelector((state) => state.courtSpecData.designsData);
   const designNames = useStoreSelector((state) => state.designName.nameList);
   const cancelRef = useRef(null);
-  const [nameCheck, setNameCheck] = useState<string>("exsited");
+  const [nameCheck, setNameCheck] = useState<string>("");
```
tatoMa, line 43: If it has an initial value as a string, you don't need to define the type of string anymore, unless you are unsure(maybe need another type other than string later)

src/components/TopBar/SaveBoard.tsx
```diff
@@ -201,6 +174,7 @@ const SaveBoard: React.FC = () => {
           h="44px"
           justifyContent="left"
           onClick={open}
+          data-testid="sava-as"
         >
           <AlertDialog
```
tatoMa, line 179: Please, consider extract the whole alert window to a new file.

src/components/NavBar/EditorDesignName.tsx
```diff
@@ -36,6 +40,21 @@ const DesignName = () => {
...
+      const errorMessage = checkDesignName(newDesignName.trim(), nameList);
+      if (errorMessage) {
+        setFeedbackModalOpen(true);
+        setFeedback(errorMessage);
+      } else {
+        dispatch(changeDesignName(newDesignName.trim()));
+      }
+    }
+  };
+
```
ZhihanChenZack, line 57: The name length check seems like part of the check design name function. It seems like the length check can be moved into the checkDesignName().

ZhihanChenZack (commented):
Well done!

src/components/NavBar/EditorDesignName.tsx
```diff
@@ -36,6 +40,21 @@ const DesignName = () => {
     );
   };
 
+  const saveNameChange = () => {
+    if (newDesignName.trim().length > DESIGN_NAME_MAX_CHAR_LENGTH) {
```
ZhihanChenZack, line 44: The name length check seems like part of the check design name function. It seems like the length check can be moved into the checkDesignName().

GLC-coder (approved):
Well done

PR #222, opened 2022-10-25
Feature/cc 59 order generation

wwwseven:
**1. Update shopping cart page**

- add Multi-select checkbox
- add callback function to the button
- disable the button if no selection

**2. Add order generation page**

- display order items
- create order when click the checkout button
- back to shopping cart when click the back button
- disable the checkout button if the checkbox is not checked

**3. Create order Api and order slice**

**4. Add tests for order generation**

cc59-1

cc59-2

### DIFF
--- a/src/components/OrderGeneration/OrderContainer.tsx
+++ b/src/components/OrderGeneration/OrderContainer.tsx
@@ -19,11 +19,11 @@ const OrderContainer = () => {
       quotation: item.quotation,
       quotationDetails: item.quotationDetails,
       image: item.image,
-      constructionDrawing: "todo",
+      constructionDrawing: "https://developer.mozilla.org",
     };
     return orderItem;
   });
-  const newOrder = { user_id: userId, status: "unpaid", items, depositRatio: 0.02 };
+  const newOrder = { user_id: userId, items, depositRatio: 0.02 };
 
   const handleProceedToCheckOut = () => createOrder(newOrder);
 
--- a/src/components/OrderGeneration/OrderContainer.tsx
+++ b/src/components/OrderGeneration/OrderContainer.tsx
@@ -1,0 +1,89 @@
+import { useStoreSelector } from "@/store/hooks";
+import { Flex, Text, Checkbox, Button, Link } from "@chakra-ui/react";
+import React, { useState } from "react";
+import OrderItem from "./OrderItem";
+import { switchOrderGeneration } from "@/store/reducer/buttonToggleSlice";
+import { useDispatch } from "react-redux";
+import { useCreateOrderMutation } from "@/redux/api/orderApi";
+import { depositRatio } from "@/constants/depositRatio";
+
+const OrderContainer = () => {
+  const dispatch = useDispatch();
+  const orderItems = useStoreSelector((state) => state.order);
+  const userId = useStoreSelector((state) => state.user.userId);
+  const [isChecked, setIschecked] = useState(false);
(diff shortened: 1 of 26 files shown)


src/components/ShoppingCart/CartListItem.tsx
```diff
@@ -10,27 +10,48 @@ import Image from "next/image";
...
     quotation: quotation,
     image: image,
     isExpired,
   } = item;
 
+  const handleCheckBox = (e: { target: { checked: boolean } }) => {
+    const newCheckedItems = [...checkedItems];
+    newCheckedItems[index] = e.target.checked;
+    setCheckedItems(newCheckedItems);
+  };
```
Double-Twelve, line 36: it can be replaced with
```
const handleCheckBox = () => {
  setCheckedItems(
    checkedItems.map((item, i) => {
      return index === i ? !item : item;
    })
  );
};
```

src/constants/depositRatio.ts
```diff
@@ -0,0 +1 @@
+export const depositRatio = 0.02;
```
Double-Twelve, line 1: this should be set on the admin side and then passed to the frontend, not write hard code on the frontend, because it is not easy to change later, maybe you can discuss it with BA

src/components/ShoppingCart/ShoppingCartContainer.tsx
```diff
@@ -37,6 +42,16 @@ const ShoppingCartContainer = ({ shoppingCart }: userCartList) => {
   const [cartItemIdToDelete, setCartItemIdToDelete] = useState("");
   const anyExpired = shoppingCart.some((item) => item.isExpired);
 
+  const [checkedItems, setCheckedItems] = useState(shoppingCart.map(() => false));
+  const allChecked = checkedItems.every(Boolean);
+  const isIndeterminate = checkedItems.some(Boolean) && !allChecked;
+
+  const handleCreateOrder = () => {
+    const orders = shoppingCart.filter((e, index) => checkedItems[index] === true);
```
Double-Twelve, line 50: it can be replaced with 
_const orders = shoppingCart.filter((e, index) => checkedItems[index]);_

src/components/OrderGeneration/OrderContainer.tsx
```diff
@@ -0,0 +1,90 @@
...
+        <Text fontSize="32px" fontWeight="700" textAlign="center">
+          Order Generation
+        </Text>
+        {orderItems.map((item, index) => (
+          <OrderItem key={item.id} item={item} index={index} />
+        ))}
+        <Checkbox
+          fontSize="20px"
+          fontWeight="500px"
+          letterSpacing="1px"
```
Double-Twelve, line 61: 1. _fontWeight_ should be 500 not 500px
2. why need to add _letterSpacing_?

tatoMa:
From the UX perspective, we might need a hint to tell the user why the order button is not clickable.
image

tatoMa:
Still no responsive for iPad.
image

wwwseven:
> Still no responsive for iPad. image

Will fix this in a new card.

wwwseven:
> From the UX perspective, we might need a hint or at least a reason to tell the user why the order button is not clickable. image

Solved.

Double-Twelve (approved):
Very well done Seven!

GLC-coder (approved):
Good job! LGTM

PR #223, opened 2022-10-26
add preview image to share template page

STY1997:
As the court screenshot feature has already been completed, the court preview image has been added to the create template page and image link can be stored in the database. 
Screen Shot 2022-10-26 at 21 24 53

### DIFF
--- a/src/components/CreateTemplate/index.tsx
+++ b/src/components/CreateTemplate/index.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable require-jsdoc */
 import { useStoreSelector } from "@/store/hooks";
+import Image from "next/image";
 import {
   Button,
   Modal,
@@ -20,6 +21,7 @@ import {
   useDisclosure,
   FormHelperText,
   FormErrorMessage,
+  useToast,
 } from "@chakra-ui/react";
 import React, { useRef, useState } from "react";
 import { FaUserCircle } from "react-icons/fa";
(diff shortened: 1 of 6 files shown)


src/components/CreateTemplate/index.tsx
```diff
@@ -90,12 +108,19 @@ function CreateTemplate({ isOpen, onClose }: Props) {
       }));
       return;
     }
+    if (!courtDataUrl) {
+      const title = `Fail to get courtDataUrl`;
```
Double-Twelve, line 112: Customers may not know what courtDataUrl is, maybe it's better to change it to "Fail to get court image"

src/interfaces/design.d.ts
```diff
@@ -43,6 +43,7 @@ export interface ICourtColor {
 }
 
 export interface ISaveDesign {
+  designer?: string;
```
Double-Twelve, line 46: what is case that _designer_ is undefined?

src/utils/manageExternalImage.ts
```diff
@@ -1,3 +1,9 @@
+export const generateImageFromDataUrl = async (courtDataUrl: string) => {
+  const base64 = await fetch(courtDataUrl);
+  const blob = await base64.blob();
+  return new File([blob], "default", { type: "image/png" });
+};
```
Double-Twelve, line 5: does this method need to export?

src/interfaces/design.d.ts
STY1997: > what is case that _designer_ is undefined?

This interface will be called in both creating template and generating shopping cart item. And the designer field will be only used to create the new template.

PR #225, opened 2022-10-27
Fix/cc 0112 add court preview image

STY1997:
As the court screenshot feature has already been completed, the court preview image can be added to the create template page and the image link can be saved in the database.


GLC-coder (approved):
Good Job! TQL

PR #235, opened 2022-11-03
Cc 9999 next performance in aws test

RedRe4per:
**This is not an official branch. This is a temporary code used for uat environment testing and will be rolled back as soon as possible after testing.**
The problem to test is that we found the team page to load very slowly in a uat environment. Now we want to test if the cause of this problem is "carousel"? So we used the control variable method, mirrored the team page, and removed the "carousel" component.


GLC-coder (approved):
lgtm

wwwseven (approved):

PR #249, opened 2022-11-14
Cc 110 resize and drag uploaded badge image

RedRe4per:
Feature:
1. Completed image editing component. It can be used in both cc-57, cc-110, cc-131.
2. This card does not include the "Center Badge icon on the design page". This is because the distribution of cards before is unreasonable, and the above three cards (cc-57, cc-110, cc-131.) will be done by me. Therefore, I first set the entry for editing pictures at main page => my account => edit avatar. 
3. This pull request includes functions of dragging and uploading pictures, clicking to upload pictures, picture dragging, zooming in and out of pictures, adjusting the rotation, resetting, and submitting edited pictures.
4. If you want to check the result of image editing (which is new image), pls check "console" and find the result url. The reason why this result is not rendered on the page is that firstly, it is not included in this card, and secondly, its implementation needs to modify a lot of old code, which takes a lot of time.

-------------------------------------------------------------


Cecilia0109:
The demo video is so cool!

src/components/ImageEditing/ImageEditingContainer.tsx
```diff
@@ -0,0 +1,230 @@
...
+  Slider,
+  SliderTrack,
+  SliderFilledTrack,
+  SliderThumb,
+  useToast,
+} from "@chakra-ui/react";
+import Cropper from "react-easy-crop";
+import { Point, Area } from "react-easy-crop/types";
+import { toBase64, upLoadScreenshot, inputImageCheck } from "@/utils/manageExternalImage";
+import { getCroppedImg } from "@/utils/canvasUtils";
```
tatoMa, line 24: Do you think it is possible to lazy load some of these libs? Or the whole image editing container component? 

When any relatively big size component not showing on any page initially, we lazy load it,

wwwseven (commented):
3191668580585_ pic
Excellent job! Just need to change the cursor into a hand pointer on hover.

src/utils/manageExternalImage.ts
```diff
@@ -67,3 +77,59 @@ export const deleteImage = async (imageUrl: string) => {
...
+
+export const inputImageCheck = (
+  img: File,
+  minSizeInKB: number,
+  maxSizeInKB: number,
+  toast: any
+) => {
+  const allowedTypes = [".jpg", ".png", ".jpeg"];
+  const imgPath = img.name;
+  const imgFormat = imgPath.substring(imgPath.indexOf("."));
```
Double-Twelve, line 97: 1. if the file name contains more than one ".", the format is incorrect
please use _img.type_

2. if I have a .exe file and I change the extension to .png and then upload, will get this error

image

src/components/ImageEditing/ImageEditingContainer.tsx
```diff
@@ -0,0 +1,243 @@
...
+      process.env.NEXT_PUBLIC_AVATAR_ALBUM_NAME
+    );
+    localStorage.setItem(
+      "UserInfo",
+      JSON.stringify({ ...userInfo, profileImgUrl: uploadedImageUrl })
+    );
+    dispatch(updateUserInfo({ ...userInfo, profileImgUrl: uploadedImageUrl }));
+    updateUser({ ...userInfo, profileImgUrl: uploadedImageUrl });
+    closeWindow();
+  }, [picture, croppedAreaPixels, rotation]);
```
tatoMa, line 95: You have a lot of user image specific logic here inside of this component but this file calls ImageEditingContainer? Or should it name userAvatarEditingContainer? If you want to reuse this component to fulfill any future needs (such as the image upload component for court), I believe this may need a restructure.

Howard-Cui (approved):

PR #254, opened 2022-11-21
Bug/cc 0152 fix save as and save not working on saveboard

GLC-coder:
bug/cc-0152 fix save as and save not working on saveboard is ready to be merged into main branch


wwwseven (approved):

Evezzm (approved):
LGTM

PR #273, opened 2023-04-25
Fix: DOM rendering issues in BasketballCourt compoment

Blackmesa-Canteen:
# Overview
Fixed DOM rendering issues in BasketballCourt: Console errors: `Konva has no node with the type div`.
Now the home page should be free from such kind of errors.

# Extra
It is suggested to use Html component from [react-konva-utils](https://github.com/konvajs/react-konva-utils). But the library requires a higher React.js version.


Blackmesa-Canteen:
Closed: name convention

PR #277, opened 2023-05-10
Fix: Design folder court index mismatch, meaningless code to be removed

Mason-Yao:
The design in the folder did not match its index and courtId when users tried to save an updated design due to useState/setState async issue. useDesignName would not get updated in the current render as setDesignName is async. Alternatively, it would get the value of the last render. As a result, the index and courtId would alway be the ones user previously selected.  
Before fix:
https://www.loom.com/share/375cbcb28a5841d3b354015ccc9de4fc
After fix:
https://www.loom.com/share/46e504b40d7148c08ac0e75b6489207e

### DIFF
--- a/src/components/FullCourt/index.tsx
+++ b/src/components/FullCourt/index.tsx
@@ -35,7 +35,6 @@ const FullCourt = () => {
   const canvasStates = canvasControl.canvasStates;
 
   useEffect(() => {
-    console.log(canvasStates.selectedColor);
     ref.current.x(0);
     ref.current.y(0);
   }, [canvasStates.resetState]);
--- a/src/layouts/Header.tsx
+++ b/src/layouts/Header.tsx
@@ -7,7 +7,6 @@ const Header = () => {
       <Head>
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <meta name="description" content="CourtCanva" />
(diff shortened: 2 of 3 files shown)


src/components/FullCourt/index.tsx
```diff
@@ -35,7 +35,6 @@ const FullCourt = () => {
   const canvasStates = canvasControl.canvasStates;
 
   useEffect(() => {
-    console.log(canvasStates.selectedColor);
```
Charles-Xing, line 38: That's a great approach, but I feel it's unrelated to the content of this card. Perhaps we could create a separate card later to address these irrelevant pieces of code.

src/layouts/Header.tsx
```diff
@@ -7,7 +7,6 @@ const Header = () => {
       <Head>
         <meta name="viewport" content="width=device-width, initial-scale=1" />
         <meta name="description" content="CourtCanva" />
-        <link rel="icon" href="/favicon.ico" />
```
Charles-Xing, line 10: That's a great approach, but I feel it's unrelated to the content of this card. Perhaps we could create a separate card later to address these irrelevant pieces of code.

Charles-Xing (approved):

billwisebai (approved):

PR #287, opened 2023-06-14
Fix/cc 38 rectify two variables in tile number calculation method, remove a meaningless console.log

Mason-Yao:


Charles-Xing (approved):
LGTM

Blackmesa-Canteen (approved):
fine by me

PR #297, opened 2023-06-25
feat: add spinner while waiting for download and prevent multiple cli…

Howard-Cui:
add spinner while waiting for download and prevent multiple click in a certain time interval


billwisebai (approved):

Mason-Yao (approved):
LGTM

PR #305, opened 2023-07-03
Feature/cc 34 add construction drawing

Mason-Yao:
1. Add construction drawing on design page and buttons in right-side editor bar to turn on/off the construction drawing.
image
image
2. Add x, y coordinates on the border and cartesian coordinates on each tile as the unique id.
3. Generate construction drawing image while adding to cart. An overlay is displayed which will disable all user operations. 
image
4. Add download button on orders page. A pdf file with high-resolution construction drawing will be downloaded when users click the button.
image
5. Distinct font color on construction drawing will be calculated based on WCAG 2 standard to achieve highest luminance ratio. 
https://www.w3.org/WAI/GL/wiki/Relative_luminance
https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html
image
image


src/interfaces/cartItem.d.ts
```diff
@@ -9,4 +9,5 @@ export interface ICartItem {
   quotationDetails: PriceBar[];
   image: string;
   isExpired: boolean;
+  constructionDrawing?: string;
```
tatoMa, line 12: Is this a URL string? If yes, all the same naming of this variable/type/interface is not suitable.

src/components/AddingToCartOverlay/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <Box width="100px">
+          <Spinner size="xl" color="teal.300" />
+        </Box>
+        <Text fontSize="24" color="teal.300">
+          {count > 8
+            ? "Generating design"
+            : count > 6
+            ? "Generating construction"
+            : "Adding to cart"}
+        </Text>
```
tatoMa, line 53: So if Im not wrong, this part mimics a countdown process on the screen to show users the progress of the adding cart function. But why doing this? Or is this function mentioned on the Jira board? 
I mean if it is truly necessary to make an overlay, there are components from the chakra UI right? (maybe Modal component?)

src/components/Construction/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+}
+
+const Construction = () => {
+  const dispatch = useDispatch();
+  const isConstructionOpen = useStoreSelector((state) => state.buttonToggle.isConstructionOpen);
+  const pdfRef = useRef<Konva.Stage>(null);
+  const imgSrc = useStoreSelector((state) => state.construction.constructionSrc);
+  const constructionInfo = useStoreSelector((state) => state.construction.constructionInfo);
+  const tiles = useStoreSelector((state) => state.tile.present.court);
+  const tilesColor = tiles.map((item) => item.color);
```
tatoMa, line 36: It can be memories by using React hooks such as useCallback or useMemo to reduce the compute when rerender. Same with some other methods and variables on this file.

src/components/Construction/index.tsx
```diff
@@ -0,0 +1,275 @@
...
+
+  const [img, setImg] = useState(new window.Image());
+  useEffect(() => {
+    if (imgSrc) {
+      const loadImg = new window.Image();
+      loadImg.onload = () => {
+        setImg(loadImg);
+      };
+      loadImg.src = imgSrc;
+    }
```
tatoMa, line 62: if imgSrc
  img.src = imgSrc


Anyway, this part confuses me.

src/components/MyOrder/MyOrderItem.tsx
```diff
@@ -1,12 +1,36 @@
-import { Flex, Box, Text, Stack, Badge } from "@chakra-ui/react";
+import { Flex, Box, Text, Stack, Badge, Button } from "@chakra-ui/react";
 import Image from "next/image";
 import formatCurrency from "@/utils/formatCurrency";
+import { downloadToPDF } from "@/utils/printConstruction";
```
tatoMa, line 4: How about try lazy loading?

src/components/AddingToCartOverlay/index.tsx
Mason-Yao: For the first question, we need to disable all operations that may modify the design when add-to-cart is processing. Any changes made to the design during this period will trigger a new add-to-cart process, which is not what we want. The countdown information is just for making the entire process more user-friendly. 
For the second question, I tried Modal but it seems it will be a little bit jerking due to its rendering machenism. My thought is an overlay also makes sense as we only need an overlay rathan than a modal with other components.

src/interfaces/cartItem.d.ts
Mason-Yao: Yes I agree. I did it to align the naming on the backend. I initally used a more readable name but found that I would have to change all the relevant naming on the backend, so I use the same name. What should I do in this situation?

src/components/Construction/index.tsx
Mason-Yao: The main reason for dealing with it like this is because Konva needs a fully loaded image.
The process of drawing image in Konva is like following:
1. Create a HTMLImageElement with `constloadImg = new window.Image();`
2. In Konva, `<Image image={img} />`, Image will draw a canvas with the image at the moment of rendering. 
3. When we set `loadImg.src = imgSrc`, loadImg will begin to load asynchronourly. At this moment, image is still blank.
4. We use `loadImg.onload = () => {setImg(loadImg);};` to listen to the load event, set image to loadImg when it completes

src/components/MyOrder/MyOrderItem.tsx
Mason-Yao: Do you mean we can try lazy loading for items in the order list and cart list? I agree with that. Should we use a new card to add this new feature?

src/components/Construction/index.tsx
Mason-Yao: Some funcs and variables has been rewritten with useCallback and useMemo.

src/components/MyOrder/MyOrderItem.tsx
tatoMa: We don't need a new card for this. Please look at our existing coding, there are lazy loading imports somewhere.

PR #318, opened 2023-08-08
Feat/cc 54 quotation expiration

Kaifeng-Sun:
1. The system should accurately calculate the duration from the current date according to the default expiration rules. On my order page, the expiration date should be displayed to the user: e.g. "Quotation will expire in 21 days."  
image
image
2. If the quotation exceed the expired date, the system should prevent any further actions or modifications on an expired quotation. The order status should display "Quotation expired". 
image
image


Mason-Yao (approved):
LGTM

PR #320, opened 2023-08-18
Feat/cc69 save image from backend to s3 frontend

billwisebai:
1. Send the image in Base64 to the backend using designApi
2. Update the SaveBoard with the new method
3. Fix the error of unauthorised to get design data


Howard-Cui (approved):

Mason-Yao (approved):
LGTM

PR #322, opened 2023-08-22
fix: fix the problem that cannot reload when modify orders

Howard-Cui:


Kaifeng-Sun (approved):
LGTM

src/pages/my_order.tsx
```diff
@@ -68,18 +70,20 @@ const MyOrder = () => {
...
-    <ProfileItemContainer>
-      <MyOrderContainer myOrders={myOrders} />
-      {imgSrc && courtDesign && <PdfCanvas courtDesign={courtDesign} imgSrc={imgSrc as string} />}
-    </ProfileItemContainer>
+    <updateOrderContext.Provider value={[updateSignal, setUpdateSignal]}>
+      <ProfileItemContainer>
+        <MyOrderContainer myOrders={myOrders} />
+        {imgSrc && courtDesign && <PdfCanvas courtDesign={courtDesign} imgSrc={imgSrc as string} />}
+      </ProfileItemContainer>
+    </updateOrderContext.Provider>
```
tatoMa, line 86: So you wrap some states in a context provider. Have you noticed that there is no context used anywhere in the CC app? We have Redux, we can pass state as props, but we don't have context now.

Unless you are sure what you doing, I cannot suggest you use context here.

tatoMa (approved):
LGTM

